PR #18, opened 2017-10-11
Add logic for managing users and groups in PVE

lae:
This adds `proxmox_user` and `proxmox_group` modules for creating and managing users and groups in PVE. It is cluster-aware and will only run once per cluster if being ran with `pve_cluster_enabled` set to yes/True.

These are defined through the role variables `pve_groups` and `pve_users`.

### DIFF
--- a/library/proxmox_group.py
+++ b/library/proxmox_group.py
@@ -14,12 +14,20 @@ short_description: Manages groups in Proxmox
 
 options:
     name:
-        description:
-            - Group name
         required: true
-    comment:
+        aliases: [ "group", "groupid" ]
         description:
-            - Group comment
+            - Name of the PVE group to manager.
+    state:
+        required: false
+        default: "present"
+        choices: [ "present", "absent" ]
+        description:
+            - Specifies whether the group should exist or not.
+    comment:
+        required: false
+        description:
+            - Optionally sets the group's comment in PVE.
 
 author:
     - Musee Ullah (@lae)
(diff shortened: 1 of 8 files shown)


library/proxmox_group.py
```diff
@@ -0,0 +1,154 @@
...
+        try:
+            self.cluster.access.groups(self.name).put(**updated_group)
+            return (True, None)
+        except:
+            return (False, "Failed to run pvesh create for this group.")
+
+def main():
+    # Refer to https://pve.proxmox.com/pve-docs/api-viewer/index.html
+    module = AnsibleModule(
+        argument_spec = dict(
```
tankbusta: Not sure if you're aware but using the dictionary literal, `{}` is faster. Obviously not a huge issue as it's not part of a tight loop. 

Source: https://stackoverflow.com/questions/6610606/is-there-a-difference-between-using-a-dict-literal-and-a-dict-constructor

lae: Going to leave this one as-is since it seems to be standard practice in the core Ansible modules, maybe for readability? But TIL, thanks.

library/proxmox_group.py
```diff
@@ -0,0 +1,154 @@
...
+        self.state = module.params['state']
+        self.comment = module.params['comment']
+
+        self.cluster = ProxmoxAPI(backend='local')
+
+    def group_exists(self):
+        try:
+            if self.cluster.access.groups.get(self.name):
+                return True
+        except:
```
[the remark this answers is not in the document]
lae: `.get()` here isn't actually a dict and will return an exception if it doesn't exist. It's provided by the `proxmoxer` library and basically just shells out `pvesh get /access/groups/{groupid}` - I'm leaving removing/modifying this dependency as a future task for now.

PR #25, opened 2018-03-13
Allow upgrade step to be configurable

roadrunner:
sometimes dist upgrade operation can be dangerous.
so would nice to set otherwise.
ps: it's enabled by default

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -65,7 +65,7 @@
   apt:
     update_cache: yes
     cache_valid_time: 3600
-    upgrade: pve_upgrade
+    upgrade: "{{ pve_upgrade }}"
 
 - include: identify_needed_packages.yml
 
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -66,6 +66,7 @@
     update_cache: yes
     cache_valid_time: 3600
     upgrade: dist
+  when: pve_dist_upgrade | default(true)
 
 - include: identify_needed_packages.yml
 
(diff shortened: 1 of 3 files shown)


tasks/main.yml
```diff
@@ -66,6 +66,7 @@
     update_cache: yes
     cache_valid_time: 3600
     upgrade: dist
+  when: pve_dist_upgrade | default(true)
```
lae: The `default` filter here is redundant since it's already specified as a default in `defaults/main.yml`. Can you remove this?

lae: ~The `default` filter here is redundant since it's already specified as a default in `defaults/main.yml`. Can you remove this?~

Actually, I think it would be a better option to move the variable to the `upgrade` parameter and allow users to configure either a safe-upgrade or dist-upgrade, because with this method we're skipping all upgrades.

lae:
Can you also add this variable to the documentation as well?

tasks/main.yml
roadrunner: you're right, with upgrade we have better options to perform upgrade.

tasks/main.yml
```diff
@@ -65,7 +65,7 @@
   apt:
     update_cache: yes
     cache_valid_time: 3600
-    upgrade: dist
+    upgrade: pve_upgrade
```
lae: `pve_upgrade` needs to be `{{ pve_upgrade }}` so that the variable gets interpreted properly

roadrunner: sorry, fixed.

PR #26, opened 2018-03-26
Limit root login between PVE cluster hosts to key authentication

mejo-:
To my knowledge, there's no need to allow password authentication.

Btw, thanks for the great role. I'm just starting to use it :)

### DIFF
--- a/tasks/ssh_cluster_config.yml
+++ b/tasks/ssh_cluster_config.yml
@@ -44,7 +44,7 @@
     content: |
       {% for host in groups[pve_group] %}
       Match Address {{ hostvars[host].ansible_default_ipv4.address }}
-      PermitRootLogin yes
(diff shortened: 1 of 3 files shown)


lae:
Role variable for this seems very unnecessary. SSH keys are setup across all hosts and this line only affects inter-host connectivity, so just hardcoding `prohibit-password` is the simplest solution and doesn't break anything.

mejo-:
The reason for a role variable was that the name changed in OpenSSH between Jessie and Stretch. In OpenSSH 7.4 (in Debian Stretch), `prohibit-password` seems to be the preferred wording.

But I can change my PF if you prefer to use `without-password` :)

lae:
Oh I misread where the variable was stored, for some reason I thought it was in defaults (no wonder why I typed `prohibit-password` first when I typed my reply and then edited it later). Alright, this is fine as-is then.

PR #27, opened 2018-03-26
Only run 'apt dist-upgrade' after adding the proxmox repository

mejo-:
I don't see a reason to run 'apt dist-upgrade' every time the role is
executed. It's better to do this only directly after the proxmox
repository got added/changed.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -60,12 +60,14 @@
     repo: "{{ pve_repository_line }}"
     filename: proxmox
     state: present
+  register: __pve_apt_add_repo
 
 - name: Run dist-upgrade
   apt:
     update_cache: yes
     cache_valid_time: 3600
     upgrade: "{{ pve_upgrade }}"
+  when: __pve_apt_add_repo.changed
 
 - include: identify_needed_packages.yml
 


lae:
This'll break workflows where it's expected that this role keeps PVE up to date (one of my main use cases). Setting up the Proxmox repo is a one time thing, but PVE minor releases come pretty regularly.

#25 was merged to allow users to configure whether or not they want to run the upgrade step or not (although now that I look at it, the task name should have been updated).

I guess a better method might be to:
* do an initial `dist-upgrade` when the repository is added/changed (remove the `upgrade` parameter)
* add another task that does a regular apt upgrade, but have it be configurable. whether or not to keep the `pve_upgrade` role variable for this or rename it to something else in order to not conflict with users of the existing variable (it was introduced very recently so there shouldn't be many users, if there is anyone other than @roadrunner)

Thoughts?

roadrunner:
i agree with @lae 
after #25  it's possible to enable / disable dist upgrade.
and there's nothing wrong change the `pve_upgrade` variable and re-run after the initial setup.

> pve_upgrade: dist # Performs system upgrade. Can be one of "no", "yes", "safe", "full", "dist".

ikus060:
To give, my two cent: I'm not using this role to keep proxmox updated. I'm using another playbook to keep all my server updated. The playbook is more robust in this regard.

mejo-:
Ok, I see. Would you at least accept an addition to PR #25 that limits the whole dist-upgrade task?:

```
- name: Run dist-upgrade
[...]
  when: pve_upgrade
```

That would prevent the whole task from running when `pve_upgrade: no`. My main problem with the current approach is, that the task updates the apt cache in any case, which breaks idempotency.

Besides, my personal opinion is that roles should not be used to keep systems up-to-date regarding package upgrades. Do people really use a proxmox role to install their security updates? I would rather use a dedicated playbook or other means (apt-dater, unattended-upgrades, ...) for that.

lae:
The original PR was actually like that, but I requested it be moved to the upgrade module param.

The method I suggested in my previous comment should suffice, I think. Basically, two tasks - one for install and one for optional continuous deployment. If you can update this PR with this, I'll merge it.

Regarding roles keeping systems up-to-date - this role I think fits into that niche case where it should be the *only* application role deployed (e.g. you're not going to deploy PostgreSQL on these servers...I would hope) and also manages the system's kernel packages (and identifies whether a restart is needed).

lae:
Slightly different from what I originally specified but - I pushed a branch that should work for everyone and retain idempotency by default. Waiting for tests to complete and I'll merge it+cut a release shortly after.

mejo-:
Your solution looks good to me. Thanks a lot :)

Issue #33, opened 2018-05-25
Remove support for PVE 4.x


lae:
PVE 4.x will be going EOL next month.

https://forum.proxmox.com/threads/proxmox-ve-support-lifecycle.35755/

Support for PVE 4.x will be removed at around the same time (anyone still on PVE 4.x  can still pin a specific version of this role (which you should be doing anyway) in their role requirements that has support for 4.x, which is 1.4.*).

PR #38, opened 2018-09-05
Update pvesh module_util to reflect upstream changes

SniperCZE:
There was a little change in output of pvesh in august (no 200 status line in result when OK + default json export changed to pretty tables) which needs to be reflected in pvesh ansible module.

### DIFF
--- a/module_utils/pvesh.py
+++ b/module_utils/pvesh.py
@@ -21,7 +21,8 @@ def run_command(handler, resource, **params):
     command = [
         "/usr/bin/pvesh",
         handler,
-        resource]
+        resource,
+        "--output=json"]
     for parameter, value in params.iteritems():
         command += ["-{}".format(parameter), "{}".format(value)]
 
--- a/module_utils/pvesh.py
+++ b/module_utils/pvesh.py
@@ -30,7 +30,7 @@ def run_command(handler, resource, **params):
     (result, stderr) = pipe.communicate()
     stderr = stderr.splitlines()
 
-    if stderr[0] == "200 OK":
+    if len(stderr) == 0:
         if not result:
             return {u"status": 200}
 


HenrikWMG:
Just wanna say, that i applied that Pull Request on my setup. it solved my Issues. Ty

lae:
Some tests are failing for PVE 4.x it looks like, but I have #33 open with regards to that.

PVE 5.x tests still seem to be failing here however: https://travis-ci.org/lae/ansible-role-proxmox/jobs/424784386#L1594

If you can take a look at that and fix that, I'll merge this - otherwise it might still be a bit before I can take a closer look at what might be causing that failure.

lae:
```
root@proxmox-test:~# /usr/bin/pvesh get access/users/root@pam --output=json
Use of uninitialized value in string eq at /usr/share/perl5/PVE/CLIFormatter.pm line 390.
{"enable":1,"groups":[]}
root@proxmox-test:~# /usr/bin/pvesh get access/users/root@pam --output=json 2>/dev/null
{"enable":1,"groups":[]}
```

lae:
someone filed a bug upstream that's most likely related (same error, same line) https://bugzilla.proxmox.com/show_bug.cgi?id=1914

`    return if $result_schema->{type} eq 'null';` from /usr/share/perl5/PVE/CLIFormatter.pm

lae:
Added comment to the bug and the small hotfix https://bugzilla.proxmox.com/show_bug.cgi?id=1914#c1

[Patch seems to have worked](https://travis-ci.org/lae/ansible-role-proxmox/builds/429925613), it looks like I needed to [update the tests to get JSON output from pvesh](https://github.com/lae/ansible-role-proxmox/commit/80691a77b12330ccfc163ba464c447a5f20a8c7a) as well so I'm waiting for that build to finish before merging.

PR #43, opened 2018-11-23
`with_items` is deprecated for apt module from the 2.7 version

niko2:
[DEPRECATION WARNING]: Invoking "apt" only once while using a loop via squash_actions is deprecated. Instead of using a loop to supply multiple items and specifying 
`name: {{ item }}`, please use `name: [u'linux-image-amd64', u'{{ __pve_kernel.old_packages }}']` and remove the loop. This feature will be removed in version 2.11. 
Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.

### DIFF
--- a/tasks/kernel_updates.yml
+++ b/tasks/kernel_updates.yml
@@ -24,11 +24,8 @@
 
 - name: Remove old Debian/PVE kernels
   apt:
-    name: "{{ item }}"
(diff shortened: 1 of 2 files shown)


lae:
Hi @niko2, thanks for the PR!

Can you rebase this against develop (updated Ansible versions) and squash your commits into one?

PR #44, opened 2018-11-24
with_items is deprecated for apt module from the 2.7 version

niko2:


lae:
please just force push instead of opening a new issue, then (though I think it's actually fine to not rebase it, the tests are running on latest develop)

niko2:
@lae it sounds good now

Issue #35, opened 2018-05-25
Support bare minimum storage config management


lae:
Would be a nice-to-have for a lot of people, I think.

bozho:
Ah, sorry. Not sure how I missed that one :)

Do you have any plans or ideas already? I wouldn't mind helping out.

lae:
Any ideas I've had I've forgotten by now, lol. I'll need to review the role again and experiment, but I believe it should be similar to how the other Proxmox configuration modules (see `library/proxmox_acl.py` for example) are implemented. You can also look at some closed PRs to get an idea of what all you should add (tests, docs etc).

Look at the PVE API documentation, specifically the /storage API endpoint: https://pve.proxmox.com/pve-docs/api-viewer/index.html

Think that's enough to get started? I can assign this to you if you want.

lae:
@bozho Are you still interested in tackling this?

Fbrachere:
Hi,
I'm writing this ansible storage module and i'm facing to a problem:
Once storage is created, we can not change some parameters.
For example for a "rbd" storage type, pool and username can not be changed.
I don't know what is the best strategy:
- when you play role, if storage ID exists, try to change the parameters and throw an error if they can not be changed.
- delete the storage and recreate it with updated parameters.
An ideas ?

lae:
I think deleting the storage entry might not be possible if the entry is in use by VMs?

Since Proxmox itself doesn't allow changing those parameters from the UI in the first place, I think we should throw an error if someone tries to change it in Ansible.

Fbrachere:
> I think deleting the storage entry might not be possible if the entry is in use by VMs?

Yes, it's possible even with running VMs, I tried it. But it can be dangerous, if delete is OK and there is a problem with storage creation, you're in trouble.

lae:
Another option could just be to template out /etc/pve/storage.cfg directly, but we can't use the `template` module as-is (because it breaks on the FUSE filesystem). I would just go with the simplest way and just not let the user change "immutable" params on an already existing storage ID (so yeah, throw an error).

lae:
Resolved in the PR referenced above. Thanks a lot, @Fbrachere!

@bozho Feel free to test this (`develop` branch still) and let me know if you run into any issues.

bozho:
Hi, I probably won't have the time this year, I'm quite swamped. But I'll be happy to do it in January :)

lae:
No worries!

PR #45, opened 2018-12-13
Add ability to manage proxmox storage.

Fbrachere:
As asked in #35 , a first attempt to manage storages with this role.
For now, the working type storage are:
- nfs
- dir
- rbd

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -27,4 +27,4 @@ pve_ssl_letsencrypt: false
 pve_groups: []
 pve_users: []
 pve_acls: []
-pve_storage: []
+pve_storages: []
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -27,3 +27,4 @@ pve_ssl_letsencrypt: false
 pve_groups: []
 pve_users: []
 pve_acls: []
+pve_storage: []
--- a/library/proxmox_storage.py
+++ b/library/proxmox_storage.py
@@ -22,8 +22,7 @@ options:
     type:
         required: true
         aliases: [ "storagetype" ]
-        choices: [ "zfspool", "dir", "nfs", "cifs", "glusterfs", "lvm",
-                "lvmthin", "iscsi", "iscsidirect", "rbd", "cephfs", "sheepdog", "zfs" ]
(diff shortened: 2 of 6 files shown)


defaults/main.yml
```diff
@@ -27,3 +27,4 @@ pve_ssl_letsencrypt: false
 pve_groups: []
 pve_users: []
 pve_acls: []
+pve_storage: []
```
lae: should be `pve_storages` to match the task (test is failing otherwise)

lae:
Can you add some test data in `group_vars/all` and some tests in `tests/test.yml` validating that the storages exist?

library/proxmox_storage.py
```diff
@@ -0,0 +1,335 @@
...
+    result['state'] = storage.state
+    result['changed'] = False
+
+    if storage.state == 'absent':
+        if storage.exists():
+            result['changed'] = True
+            if module.check_mode:
+                module.exit_json(**result)
+            (changed, error) = storage.remove_storage()
+
```
lae: can you remove this newline?

lae:
We can add a real NFS server to the test (see lae.nfs :) ) but I guess it wouldn't be worth it.

Just pored over this PR in more detail and overall it looks good to me + playbook run completes fine. The test tasks have the `pve_storage` typo though, which needs fixing. I'll merge afterwards.

lae:
Actually - I just noticed because of your opening PR comment. If none of the other storage types work, can we just remove them from the choices entirely, or at least comment them out, so no one tries to accidentally use them just because they're *there*?

Also sorry, I just remembered I was going to say this earlier. Can you add a `Storage` section to the documentation (`README.md`) detailing how to setup storages?

lae:
I'm restarting the failed builds, looked like a temporary networking issue or something trying to install dependencies.

lae:
@Fbrachere I'm sure you've noticed by now, but I've given you permissions to commit to this repo, so feel free to push feature branches directly to this repo/open PRs whenever you feel like it. (and if you'd rather not have that ability, let me know.)

Fbrachere:
Ok, thank you for your trust, I'll test this by adding another storage type.

PR #47, opened 2019-03-15
[WIP] Use ring0 instead of default ipv4 for hosts file and sshd configuration

gilou:
Hi,
When setting a pve_cluster_ring0_addr, the role still uses the ansible_default_ipv4.address in the hosts file for the nodes, and in the permissive sshd configuration. By default, pve_cluster_ring0_addr is set to ansible_default_ipv4.address so this won't change.
What will change will be, if you set the pve_cluster_ring0_addr:
 - the hosts file will use it
 - the sshd config file will use it

It might be better to give the user the choice, but in my opinion, that's how it should behave, if I use a ring0 addr, I probably am interested in using that address for inter node communication all the way. Also, I didn't see an obvious path to set a choice here, due to the variable priorities in ansible, but it's possible.

Thanks for this torough role by the way, it proved really useful.
Regards,
Gilou

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -29,7 +29,7 @@
     marker: "# {mark} ANSIBLE MANAGED: Proxmox Cluster Hosts"
     content: |
       {% for host in groups[pve_group] %}
-      {{ hostvars[host].ansible_default_ipv4.address }} {{ hostvars[host].ansible_fqdn }} {{ hostvars[host].ansible_hostname }}{% if ansible_fqdn == hostvars[host].ansible_fqdn %} pvelocalhost{% endif %}
+      {{ hostvars[host].pve_cluster_ring0_addr | default(hostvars[host].ansible_default_ipv4.address) }} {{ hostvars[host].ansible_fqdn }} {{ hostvars[host].ansible_hostname }}{% if ansible_fqdn == hostvars[host].ansible_fqdn %} pvelocalhost{% endif %}
 
       {% endfor %}
 
@@ -42,7 +42,7 @@
     # above, then we match against different IPs (e.g. NOT 10.0.3.17) that have
     # the hostname/fqdn we inserted a record for previously, taking care also to
     # detect word boundaries (\b wasn't working for some reason)
-    regexp: '^(?!{{ hostvars[item].ansible_default_ipv4.address | regex_escape() }} {{ hostvars[item].ansible_fqdn | regex_escape() }} {{ hostvars[item].ansible_hostname | regex_escape() }}( pvelocalhost)?)(?!{{ hostvars[item].ansible_default_ipv4.address | regex_escape() }})[\w:.]+(\s+.*)?\s({{ hostvars[item].ansible_fqdn | regex_escape() }}|{{ hostvars[item].ansible_hostname | regex_escape() }}{% if ansible_fqdn == hostvars[item].ansible_fqdn %}|pvelocalhost{% endif %})(\s+.*|\s*)$'
+    regexp: '^(?!{{ hostvars[item].pve_cluster_ring0_addr | default(hostvars[item].ansible_default_ipv4.address) | regex_escape() }} {{ hostvars[item].ansible_fqdn | regex_escape() }} {{ hostvars[item].ansible_hostname | regex_escape() }}( pvelocalhost)?)(?!{{ hostvars[item].pve_cluster_ring0_addr | regex_escape() }})[\w:.]+(\s+.*)?\s({{ hostvars[item].ansible_fqdn | regex_escape() }}|{{ hostvars[item].ansible_hostname | regex_escape() }}{% if ansible_fqdn == hostvars[item].ansible_fqdn %}|pvelocalhost{% endif %})(\s+.*|\s*)$'
     state: absent
     backup: yes
   with_items: "{{ groups[pve_group] }}"
--- a/tasks/ssh_cluster_config.yml
+++ b/tasks/ssh_cluster_config.yml
@@ -43,7 +43,7 @@
     marker: "# {mark}: Allow root logins from PVE hosts (managed by ansible)."
     content: |
       {% for host in groups[pve_group] %}
-      Match Address {{ hostvars[host].ansible_default_ipv4.address }}
+      Match Address {{ hostvars[host].pve_cluster_ring0_addr | default(hostvars[host].ansible_default_ipv4.address) }}
       PermitRootLogin prohibit-password
       {% endfor %}
     validate: "/usr/sbin/sshd -t -f %s"


lae:
build is failing on the updated task: https://travis-ci.org/lae/ansible-role-proxmox/jobs/506733208#L2903

I believe you need to change the '.ansible_default_ipv4.address` to `["pve_cluster_ring0_addr"]` (but I might be wrong). Can you make this change/squash them into one commit and force push to this branch?

gilou:
Hmm, I'm a bit dumb... being in default doesn't mean it's available as a host var / fact... Let me rework it a bit, maybe it's better to do set_fact at some point. In my use case it didn't fail because obviously this was set in my host vars...

lae:
Being in defaults should mean it's available as a hostvar. That's weird. I can't look into this into more depth (by reproducing locally) currently until I'm back in the States, but `set_fact` should not be necessary in this case.

gilou:
You can not access variables for the role from another host, it has to be a fact for that to happen (and maybe that's why the role used the default ipv4 address so as to avoid the headache)

gilou:
Or I can simply try to use ring0 and fall back to the default address using a default filter ?

lae:
Okay, after a bit more searching, I guess defaults really aren't accessible as hostvars.

Can you create a new tasks file, `load_variables.yml`, and use `set_fact` with a private `_` prefixed role var? Similar to what I have here: https://github.com/lae/ansible-role-netbox/blob/master/tasks/load_variables.yml

gilou:
OK, I squashed and force pushed using that, so that if I get the fact I use it, otherwise I switch to the `ansible_default_ipv4.address` fact.
I don't like it too much because it duplicates the defaults effort. And also, if we want to use something else, like IPv6, that's more edits to make (OK, two or three... but still.)
What about the set_fact way: we define it the same way the defaults is set, and it won't matter if it's overridden through the hostvars, as the logic will be there anyway?
```
- set_fact:
    pve_cluster_ring0_addr: "{{ pve_cluster_ring0_addr }}"
```
That will use the defaults or overridden value, and add it to hostvars ? However, in a playbook I'd do that in a previous play so that I'm sure I have the info...

lae:
Yeah, looks like you got some commits in before my last comment. The `set_fact` way is preferable, but with moving the defaults variable to a private one (mainly just to prevent some confusion as to where the var is being defined)

gilou:
I'll make a note and read on your pattern for this. The patch is wrong as it is anyway as I missed one spot in the hosts file cleanup. I'll get back on this next week most likely.

lae:
@gilou just bumping, did you get a chance to work on this?

gilou:
ah sorry, no, will try this week to get back on it

gilou:
OK.. I'm struggling.

I can indeed set a "private" _pve_cluster_ring0_addr, and use set_fact to set the fact for good (even letting the inventory / defaults take precedence, to still allow that). The goal is for hostvars to have the value if it wasn't set through the inventory.

But how do I guarantee set_fact is done for all hosts in the playgroup before I go on using it? delegate the set_fact? I'm confused.

lae:
Sorry I didn't follow up on this sooner.

I just spent a few hours (dang...but I kind of expected this) re-reading this issue/recollecting my thoughts/researching/experimenting. This is a pretty confusing issue, so to recap...

`pve_cluster_ring0_addr` is by default set to the default IPv4 address of a host in `defaults/main.yml` (`git blame` output):

```
2017-10-10 33) pve_cluster_ring0_addr: "{{ ansible_default_ipv4.address }}"
```

This makes the current commit seem redundant, since it'll always default to that (and your other `ring0_ssh` branch would be appropriate).

However, this and other role variables are only loaded for the target hosts of a playbook run, e.g. `--limit pve01`, unless they are also defined in the playbook's group_vars/host_vars. To further compound the issue, it appears that role variables are only accessible by the target host of a task: `hostvars` are not populated with role variables (https://github.com/ansible/ansible/issues/6189). This means that we can't loop through `pve_group` like we're currently doing and get `pve_cluster_ring0_addr` via `hostvars` for other hosts.

But, a `set_fact` task can fix this.

Now, I'm confused by what problem you're running into in your last comment. Are you trying to use `--limit` on maybe a single host in `pve_group`? If so, that's something that can't be supported. I can imagine that `--limit` might be attractive when adding a new host, but the other hosts in the cluster need to also have their configurations updated to include the new host. Preventing this behaviour is actually the intention of the following task in `tasks/main.yml`:

```
2018-12-14   6) - name: Ensure that we have an IP address for all cluster hosts
2018-12-14   7)   assert:
2017-03-17   8)     that:
2017-03-17   9)       - "hostvars[item].ansible_default_ipv4.address is defined"
2017-03-17  10)     msg: "Missing IP address and other information for {{ item }}. Have you gathered its facts?"
2017-10-10  11)   with_items: "{{ groups[pve_group] }}"
```

If you're not, I don't think there shouldn't be any worry that `set_fact` isn't run for all hosts before it gets used? Are you running into some other case? But if we really wanted to, delegation and `run_once` on `set_fact` may be enough, actually (and I have an example of how to use delegation below).

Since you're probably busy, I'm going to close this and open another PR with the `set_fact` solution so I can cut a release, but please respond when you get the time. (I'll get notifications on Discord.)

---

The following is just extra information about some different paths from `set_fact` I ended up exploring.

The `include_vars` module can also populate facts, it seems. So the following could work to load all the defaults for all of our other hosts:

```yaml
- name: generate role facts for other hosts
  include_vars:
    file: defaults/main.yml
  delegate_to: "{{ item }}"
  delegate_facts: True
  loop: "{{ groups['pve_group'] }}"
```

However, it completely overrides any group_vars/host_vars. Apparently there's been quite some discussion about (effectively) adding an `overwrite` parameter to this module, but it was rejected:

https://github.com/ansible/ansible/pull/23738
https://github.com/ansible/proposals/pull/21
https://meetbot.fedoraproject.org/ansible-meeting/2019-06-20/ansible_core_irc_plublic_meeting.2019-06-20-15.03.log.html
https://groups.google.com/forum/#!topic/ansible-devel/J22Sbo9p2cI

I was thinking that maybe we could include that patched module in here, but I guess that'd increase the maintenance burden (and it's already out of sync with ansible `devel`).

Another thing I tried was the [`include_role`](https://docs.ansible.com/ansible/latest/modules/include_role_module.html#include-role-module) module and using an empty tasks file, but apparently I can't delegate it:

```
ERROR! 'delegate_to' is not a valid attribute for a IncludeRole

The error appears to be in '/home/lae/tmp/playbooks/role.delegate_facts/tasks/main.yml': line 16, column 3, but may be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:
- include_role:
  ^ here
```

gilou:
damn! That's a lot of feedback and introspection!

Thanks for following up, I definitely agree on your conclusion. Regarding my issue, I missed the assert, and I agree that's enough. Supporting --limit requires a lot of work to do things that have to be done for fact gathering, and I don't require it, especially if we can't, and that's fine!

Cheers
Gilles

PR #48, opened 2019-04-01
Don't install postfix

gaudenz:
Hard-coding an MTA to install prevents users of this role from using an
alternative MTA of their choice.

While an MTA is required to use Proxmox VE there is already a dependency
on the mail-transport-agent meta-package in the Proxmox packages which
ensures that this requirement is satisfied. Leave the choice of which MTA
to use to the user and don't hard-code it in this role.

### DIFF
--- a/tasks/identify_needed_packages.yml
+++ b/tasks/identify_needed_packages.yml
@@ -3,7 +3,6 @@
   set_fact:
     _pve_install_packages:
       - proxmox-ve
-      - postfix
       - open-iscsi
       - ksm-control-daemon
       - systemd-sysv


gaudenz:
The CI error is unrelated to this change. It fails because it's unable to verify the self-signed certificate of the Proxmox Web UI.

gaudenz:
I added a fix for the test failure on Ansible 2.8

lae:
just for my reference: https://git.proxmox.com/?p=pve-manager.git;a=blob;f=debian/control;h=07ab53eb6d0239e20ecbb3bb191c35e6c63beb2a;hb=HEAD#l62

PR #49, opened 2019-04-02
Don't modprobe in Travis CI tests

gaudenz:
A change in the modprobe module in Ansible 2.8 made the module fail even
if the state is already correct. Because of this the task "Unload
nmi_watchdog" failed in tests.

Loading and unloading modules does not work when running tests on Travis
CI in an LXC container. This adds the "skiponlxc" tag to all invocations of
the modprobe Ansible module.

### DIFF
--- a/tasks/disable_nmi_watchdog.yml
+++ b/tasks/disable_nmi_watchdog.yml
@@ -3,6 +3,7 @@
   modprobe:
     name: nmi_watchdog
     state: absent
+  tags: skiponlxc
 
 - name: Attempt to unload softdog live
   modprobe:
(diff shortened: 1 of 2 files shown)


lae:
The `skiponlxc` tag is more of a last resort to get tests to pass (rather than an indication that this role should be used inside LXC containers).

The change in the modprobe module throws an exception when it can't find `/lib/modules/$(uname -r)/modules.builtin` and doesn't actually have anything to do with the execution of modprobe (whether or not they *should* catch that exception is debatable though...so I decided not to send a patch). This is fixed in b1ea4297578357026161719d6fdeba682e18eacd.

PR #50, opened 2019-04-03
HA groups

gaudenz:
Add functionality to create HA groups. This does not really depend on the Travis CI test fix, but without it the CI does not pass and the tests added in this PR won't even be run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,6 +435,22 @@ pve_datacenter_cfg:
   keyboard: en-us
 ```
 
+You can also configure [HA manager groups][ha-group]:
+```
+pve_cluster_ha_groups: [] # List of HA groups to create in PVE.
+```
+
+This example creates a group "lab_node01" for resources assigned to the
+lab-node01 host:
+```
+pve_cluster_ha_groups:
+  - name: lab_node01
+    comment: "My HA group"
+    nodes: "lab-node01"
+    nofailback: 0
+    restricted: 0
+```
+
 All configuration options supported in the datacenter.cfg file are documented in the
 [Proxmox manual datacenter.cfg section][datacenter-cfg].
 
@@ -629,3 +645,4 @@ Michael Holasek ([@mholasek](https://github.com/mholasek))
 [storage-module]: https://github.com/lae/ansible-role-proxmox/blob/master/library/proxmox_storage.py
 [datacenter-cfg]: https://pve.proxmox.com/wiki/Manual:_datacenter.cfg
 [ceph_volume]: https://github.com/ceph/ceph-ansible/blob/master/library/ceph_volume.py
+[ha-group]: https://pve.proxmox.com/wiki/High_Availability#ha_manager_groups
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -409,7 +409,22 @@ pve_cluster_ring0_addr: "{{ ansible_default_ipv4.address }}"
...
 # pve_cluster_bindnet1_addr: "{{ pve_cluster_ring1_addr }}"
+```
 
+You can also configure [HA manager groups][ha-group]:
+```
+pve_cluster_ha_groups: [] # List of HA groups to create in PVE.
+```
+
+This example create a group "lab_node01" for resources assigned to host
+lab-node01:
```
lae: create → creates
host lab-node01 → the lab-node01 host

README.md
```diff
@@ -409,7 +409,22 @@ pve_cluster_ring0_addr: "{{ ansible_default_ipv4.address }}"
...
+```
+pve_cluster_ha_groups:
+  - name: lab_node01
+    comment: "My HA group"
+    nodes: "lab-node01"
```
lae: I think this should probably be another list rather than a string, thoughts?

tests/group_vars/all
```diff
@@ -12,6 +12,10 @@ pve_zfs_zed_email: root@localhost
 pve_ssl_private_key: "{{ lookup('file', ssl_host_key_path) }}"
 pve_ssl_certificate: "{{ lookup('file', ssl_host_cert_path) }}"
 pve_cluster_enabled: yes
+pve_cluster_ha_groups:
+  - name: proxmox_5_01
+    comment: "Resources on proxmox-5-01"
+    nodes: proxmox-5-01
```
lae: and adding to the comment about making this a list - we could add the other 2 proxmox test containers here

gaudenz: OK if we make this a list just to test if having a list with more than one item actually works. I would probably add 2 containers to also test that it does not unconditionally add all of them ;-).

README.md
gaudenz: I don't have a strong opinion on this. The way I use HA groups in Proxmox, having a group assigned to multiple nodes does not make much sense. I use them to assign resources to the node where they run under normal conditions (no maintenance, no hardware failure). In the current implementation this is just an opaque string used as an argument when creating the group. This keeps the implementation simple. On the other hand having this as a list would make it more strict. In the end it's probably just a matter of adding a "join".

gaudenz:
Instead of converting this to a module I added a second tasks to update group properties if they don't match. I also changed the `run_once: True` condition to a `when:` condition and added if statements around optional command parameters as requested in the other pull requests.

gaudenz:
@lae Is this now OK or would you like more changes?

lae:
Sorry for putting this off for so long. I was still kind of thinking that the nodes value should a list (as one could make an "HA" inventory group and just pass it) but wanted to check out how it actually worked. I guess since the API returns the nodes as a comma-delimited string the current PR is fine as-is (the following is mainly for my record):

```
vagrant@pve-1:~$ sudo pvesh get /cluster/ha/groups --output=json
[{"comment":"Resources on first cluster host","digest":"d3838b51810de14d10d24475c8fb866172548190","group":"test","nodes":"pve-1","restricted":1,"type":"group"},{"comment":"test2 resources","digest":"fbee3c9903103db28cf7dddbe031e14af67d7cb7","group":"test2","nodes":"pve-1,pve-2","restricted":1,"type":"group"}]
```

PR #51, opened 2019-04-03
Configuration of datacenter.cfg

gaudenz:
Allow configuration of /etc/pve/datacenter.cfg from a dictionary.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -178,6 +178,37 @@
   with_items: "{{ pve_storages }}"
   when: "not pve_cluster_enabled or (pve_cluster_enabled and inventory_hostname == groups[pve_group][0])"
 
+- name: Check datacenter.cfg exists
+  stat:
+    path: "/etc/pve/datacenter.cfg"
+  register: _datacenter_cfg
+  when:
+    - not pve_cluster_enabled or (pve_cluster_enabled and inventory_hostname == groups[pve_group][0])
+    - pve_datacenter_cfg | length > 0
+
+- name: Create datacenter.cfg if it does not exist
+  file:
+    path: "/etc/pve/datacenter.cfg"
+    state: "touch"
+  when:
+    - not pve_cluster_enabled or (pve_cluster_enabled and inventory_hostname == groups[pve_group][0])
+    - pve_datacenter_cfg | length > 0
+    - not _datacenter_cfg.stat.exists
+
+- name: Configure datacenter.cfg
+  copy:
+    dest: "/etc/pve/datacenter.cfg"
+    owner: "root"
+    group: "www-data"
+    mode: "0640"
+    content: |
+      {% for k,v in pve_datacenter_cfg.items() %}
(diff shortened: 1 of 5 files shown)


tasks/main.yml
```diff
@@ -178,6 +178,35 @@
...
+  when: pve_datacenter_cfg | length > 0
+
+- name: Create datacenter.cfg if it does not exist
+  file:
+    path: "/etc/pve/datacenter.cfg"
+    state: "touch"
+  run_once: True
+  when:
+    - pve_datacenter_cfg | length > 0
+    - not _datacenter_cfg.stat.exists
```
lae: Is there a reason why the `stat` task and this conditional is needed? The `file` module should perform a check of whether or not it exists, I'm pretty sure...

tasks/main.yml
```diff
@@ -178,6 +178,35 @@
...
+    path: "/etc/pve/datacenter.cfg"
+  register: _datacenter_cfg
+  run_once: True
+  when: pve_datacenter_cfg | length > 0
+
+- name: Create datacenter.cfg if it does not exist
+  file:
+    path: "/etc/pve/datacenter.cfg"
+    state: "touch"
+  run_once: True
```
lae: `run_once` shouldn't be used here, otherwise in some setups we'll see aberrant behaviour - e.g. someone running a playbook deploying 5 separate (un-clustered) proxmox hosts would only see these tasks run on the first host.

tasks/main.yml
```diff
@@ -178,6 +178,35 @@
...
+- name: Configure datacenter.cfg
+  copy:
+    dest: "/etc/pve/datacenter.cfg"
+    owner: "root"
+    group: "www-data"
+    mode: "0640"
+    content: |
+      {% for k,v in pve_datacenter_cfg.items() %}
+      {{ k }}: {{ v }}
+      {% endfor %}
```
lae: This can be simplified to just
```
    content: "{{ pve_datacenter_cfg | to_yaml }}"
```
I believe.

tasks/main.yml
gaudenz: This is to ensure that the file is not "touched" (ie updating the access and modification times) if it is not changed. The `file` modules touch operation is only needed when creating the file.

tasks/main.yml
gaudenz: Yes and no. In theory the file format is not a YAML file, so no. But in practice it probably does not matter much as long as pve_datacenter_cfg is a simple flat dictionary with only strings as keys and values. There is currently no code to enforce this, so yes it can be simplified. If you prefer I can change it.

tasks/main.yml
gaudenz: Fixed.

trickert76:
Sorry to comment an open issue. Is it really a good idea to edit the file itself instead of using `pvesh`?

    pvesh get cluster/options

Returns a list of configurations. And setting it could also be done via Python module as all other mods.

Issue #57, opened 2019-06-07
Deprecated warnings in ansible 2.8 with v1.5


trickert76:
Hi - I love this project - it makes my life much more easier. But there are at least some very minor but later more major warnings. With Ansible 2.8 some variable managing changes - so this should be easy be updated to avoid later problems with newer Ansible releases

TASK [lae.proxmox : Ensure this host is in the group specified] **********************************************************************************************************
[DEPRECATION WARNING]: evaluating pve_cluster_enabled as a bare variable, this behaviour will go away and you might need to add |bool to the expression in the future. 
Also see CONDITIONAL_BARE_VARS configuration toggle.. This feature will be removed in version 2.12. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
skipping: [ac-sh002]

...

TASK [lae.proxmox : Perform system upgrades] *****************************************************************************************************************************
[DEPRECATION WARNING]: evaluating pve_run_system_upgrades as a bare variable, this behaviour will go away and you might need to add |bool to the expression in the 
future. Also see CONDITIONAL_BARE_VARS configuration toggle.. This feature will be removed in version 2.12. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
skipping: [ac-sh002]

...

TASK [lae.proxmox : Stage ZFS packages if ZFS is enabled] ****************************************************************************************************************
[DEPRECATION WARNING]: evaluating pve_zfs_enabled as a bare variable, this behaviour will go away and you might need to add |bool to the expression in the future. Also 
see CONDITIONAL_BARE_VARS configuration toggle.. This feature will be removed in version 2.12. Deprecation warnings can be disabled by setting deprecation_warnings=False
 in ansible.cfg.
skipping: [ac-sh002]

...

TASK [lae.proxmox : Remove subscription check wrapper function in web UI] ************************************************************************************************
[DEPRECATION WARNING]: evaluating pve_remove_subscription_warning as a bare variable, this behaviour will go away and you might need to add |bool to the expression in 
the future. Also see CONDITIONAL_BARE_VARS configuration toggle.. This feature will be removed in version 2.12. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
ok: [ac-sh002]

TASK [lae.proxmox : Check for kernel update] *****************************************************************************************************************************
[...]

PR #58, opened 2019-06-07
#57 changed when condition to use bool declaration for boolean variables

trickert76:
Please have a look. At the moment I didn't test that - because I don't have an idea how to integrate the git repo into my playbook instead of the ansible-galaxy role.

### DIFF
--- a/tasks/identify_needed_packages.yml
+++ b/tasks/identify_needed_packages.yml
@@ -16,7 +16,7 @@
 - name: Stage ZFS packages if ZFS is enabled
   set_fact:
     _pve_install_packages: "{{ _pve_install_packages | union(['zfsutils-linux', 'zfs-initramfs', 'zfs-zed']) }}"
-  when: pve_zfs_enabled|bool
+  when: pve_zfs_enabled | bool
 
 - name: Stage any extra packages the user has specified
   set_fact:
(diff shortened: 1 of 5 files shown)


lae:
Could you add spaces around the `|` operator (an amend commit is fine)? The existing codebase follows that style and I'd like to keep it consistent.

This is for your reference but you can install roles from github (or any other vcs source really) like below.

```bash
ansible-galaxy install $URL,$VERSION,$ROLENAME
# e.g.
ansible-galaxy install https://github.com/lae/ansible-node-exporter/archive/feature/download_url.tar.gz,download_url,lae.node-exporter --force
ansible-galaxy install https://github.com/lae/ansible-role-proxmox/archive/develop.tar.gz,v1.6.0-pre,lae.proxmox --force
```

You could also just symlink `~/.ansible/roles/lae.proxmox` to your checked out copy as well, since that's where Ansible looks for roles (and thus ansible-galaxy installs roles into by default). I actually use this method for development/local testing.

trickert76:
Yeah, thanks - I learned about the requirements.yml

```
---
- src: geerlingguy.ntp
- src: systemli.letsencrypt
- src: https://github.com/trickert76/ansible-role-proxmox.git
  version: fix/remove_warning_bare_boolean_variables
  name: lae.proxmox
```
and calling 

`ansible-galaxy install -r requirements.yml --force`

trickert76:
Just one question about the code style.
If it is
`when: var | bool`

should it be better
`when: "var | bool"`

lae:
Actually, good point. It'd be best if they were quoted, too.

trickert76:
Ok, I think, I'm ready.

PR #60, opened 2019-07-25
PVE 6 support

trickert76:
Proxmox v6 Support - in my environment it works on a fresh Debian Buster.


AleRousse:
Unfortunately i can't achieve to make this works with a debian 10 machine.
I am getting error when the role is trying to setup cluster with pvecm create.

I found that in proxmox VE 6.X you need to specify --linkX and not -bindnetX_addr and ringX_addr
You can verify this by the following: 
https://forum.proxmox.com/threads/separate-cluster-network-wiki-bindnet0_addr-command.25342/
https://pve.proxmox.com/pve-docs/pve-admin-guide.html#_cluster_network

Let me know what you think about it

lae:
Posts \#5 and \#6 are the relevant posts in the linked thread (post 5 permalink: https://forum.proxmox.com/threads/separate-cluster-network-wiki-bindnet0_addr-command.25342/#post-259509)

Reading through the above, and given the fact `bindnet*` is gone entirely, I think duplicating the initialization task (https://github.com/lae/ansible-role-proxmox/blob/147d366a05b2192ee0bd47f82c5b72ac89ed3bc8/tasks/pve_cluster_config.yml#L37) using the `link*` parameters, templated via a new role variable of the same name, with a when condition (distribution == buster) might be more maintainable. The README would then just have a note saying the bindnet/ringnet parameters are for PVE 5.x.

Thoughts?

AleRousse:
@lae Yeah I think to duplicating and making task distribution specific would be more clear. 

Also I am expenriencing another task failing in the following: 

https://github.com/trickert76/ansible-role-proxmox/blob/feature/proxmox6_support/tasks/pve_add_node.yml#L9

Again I think the solution would be duplicating the task and making it distribution specific

tasks/pve_add_node.yml
```diff
@@ -5,7 +5,7 @@
 
 - name: Add node to Proxmox cluster
   command: "pvecm add {{ hostvars[groups[pve_group][0]]['ansible_default_ipv4']['address'] }}\
-            {{ ' -use_ssh' if ansible_distribution_release == 'stretch' else '' }} \
+            {{ ' -use_ssh' if (ansible_distribution_release == 'stretch' || ansible_distribution_release == 'buster') else '' }} \
```
lae: The conditional here can be removed entirely - I missed this when removing support for 4.x.

trickert76:
Great - I'll change that too and make a new PR. I would make two tasks of it and add a when condition

lae:
@trickert76 that change should be in this PR, not a separate one (it's a required change for the role to continue working in clusters for PVE 6 as far as I can tell?)

trickert76:
I don't understand that error in Travis.

trickert76:
How could I define the distribution version in the test - I only see in tests/install.yml a profile - but it should be a second test for Buster too. Also I don't understand - if it is a Stretch test, why should it be changed, when the "when" condition is working.

lae:
@trickert76 this is a tricky one - the space before `-use_ssh` is missing now so it tries to add `{{ host_ip }}-use_ssh` as a node. I'm not sure if this is what causes timeouts for the first two, but it's visible in the last one. Try moving `-use_ssh` to the line before.

As for adding tests for PVE 6, the following steps are needed:

- add 3 new nodes to `tests/inventory`
- update `tests/install.yml` by adding another item to `test_profiles`, profile = `debian-buster` and prefix = `proxmox-6-`

But right now there isn't a debian-buster test profile in the test suite yet, so it won't work. Let me get to updating the test suite and I'll add a commit for PVE 6 tests.

lae:
@AleRousse Could I have you try to deploy this a PVE 6 cluster with this branch and report back if you have any other issues?

AleRousse:
@trickert76 and @lae  I confirm that it works well for a PVE6's cluster
Tested with the following feature:
- PVE 6 Base installation
- Storage content with LVM
- Clustering 2 nodes with Corosync 3.0.2

Thanks all!

PR #64, opened 2019-08-09
Allow SSH port for cluster hosts to be configurable

trickert76:
If the SSH port on the hosts is not the default port, then joining a Cluster with -use-ssh will never work. With the new variable pve_ssh_port it is possible to set the specific port. It will be written to /etc/ssh/ssh_config which is already changed by hostkey assignment and it makes sense to change the default port there too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 lae.proxmox
 ===========
 
-Installs and configures a Proxmox 5.x cluster with the following features:
+Installs and configures a Proxmox 5.x/6.x cluster with the following features:
 
 - Ensures all hosts can connect to one another as root
 - Ability to create/manage groups, users, access control lists and storage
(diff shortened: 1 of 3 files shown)


lae:
Can you rebase this on develop?

trickert76:
I'm not sure, if I rebased it correctly. I don't understand, why there are still old commits here - but the changed files (2) are correct.

lae:
@trickert76 rebases are different from merges. There are a lot of extra commits here that shouldn't be here, which is what a rebase on develop is intended to remove.

To clean this up, the easiest way is probably to just recreate the branch now and cherry-pick the appropriate commit instead of using `git rebase`:

```
git rename -m feature/custom_ssh_port oldsshportbranch # rename this branch so we can reuse the name, but still be able to reference the commits in here (although I'm not sure if it's necessary to still keep the branch)
git checkout -b feature/custom_ssh_port 611c3d10020c560e610582976b48aab2c1f7f3e9 # current upstream develop hash to rebase on
git cherry-pick 5c0b8ea11646f549fcba399635c575defdbd1575 # ssh role variable commit
git push --force
```

(rebasing involves overwriting history, hence a force push is necessary - and that's fine for feature branches)

Also I noticed that the README is missing an update for the new role variable, could you add that, too?

PR #75, opened 2019-10-11
Corosync role variable handling

lae:
This collapses all of the address-related variables used for bringing up clusters to `pve_cluster_addr0` and `pve_cluster_addr1`, while maintaining backwards compatibility. The bindnet parameters are no longer passed to `pvecm` either, as they turned out to be redundant to the ring parameters.

This also removes instances in the configuration where the hosts' default IPv4 address was used even if ring0 address was overridden. See #47 for more information.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -29,14 +29,8 @@ pve_ceph_crush_rules: []
 # pve_ssl_certificate: "contents of certificate"
 pve_cluster_enabled: no
 pve_cluster_clustername: "{{ pve_group }}"
-# PVE 5.x (Debian Stretch) clustering options
-pve_cluster_ring0_addr: "{{ ansible_default_ipv4.address }}"
-pve_cluster_bindnet0_addr: "{{ pve_cluster_ring0_addr }}"
-# pve_cluster_ring1_addr: "another interface's IP address or hostname"
-# pve_cluster_bindnet1_addr: "{{ pve_cluster_ring1_addr }}"
-# PVE 6.x (Debian Buster) clustering options
-pve_cluster_link0_addr: "{{ ansible_default_ipv4.address }}"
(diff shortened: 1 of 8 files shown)


lae:
Backwards compat test: https://travis-ci.org/lae/ansible-role-proxmox/jobs/596410730

lae:
@trickert76 if you've got time to look at this, let me know. Otherwise I'm just gonna merge near the end of the week.

trickert76:
I'll try to check that today.

trickert76:
Looks good to me.

defaults/main.yml
```diff
@@ -29,14 +29,8 @@ pve_ceph_crush_rules: []
...
 pve_cluster_clustername: "{{ pve_group }}"
-# PVE 5.x (Debian Stretch) clustering options
-pve_cluster_ring0_addr: "{{ ansible_default_ipv4.address }}"
-pve_cluster_bindnet0_addr: "{{ pve_cluster_ring0_addr }}"
-# pve_cluster_ring1_addr: "another interface's IP address or hostname"
-# pve_cluster_bindnet1_addr: "{{ pve_cluster_ring1_addr }}"
-# PVE 6.x (Debian Buster) clustering options
-pve_cluster_link0_addr: "{{ ansible_default_ipv4.address }}"
-# pve_cluster_link1_addr: "another interface's IP address or hostname"
+# pve_cluster_addr0: "{{ ansible_default_ipv4.address }}"
```
trickert76: Just my two cents, I've learned, that the default interfaces may change. For example docker installations add docker0 and then because d is before e, the default interfaces in ansible changes. Maybe we should mention that. I've added somewhere else in may playbook a set_fact based on a filter of all interfaces starting with 'e' to get the first address. But when you use more then one ethernet device, this will maybe overridden. Nothing to change here.

PR #76, opened 2019-10-11
Add multi-machine Vagrantfile/playbook [ci skip]

lae:


lae:
Accidentally committed the py3 fixes to this branch, but whatever.

Tests passed here: https://travis-ci.org/lae/ansible-role-proxmox/builds/596548597

Fails on 2.6.x because of a bug: https://github.com/ansible/ansible/issues/51665. Since I don't really think it matters much to test for py3 currently, going to leave the tests on py2 (so I don't have to bump the min required version yet).

PR #81, opened 2019-11-17
Identify Ceph OSDs correctly to maintain idempotency

mholasek:
Added idempodency for osd creation.

1. Read existing OSDs via the ceph_volume module
2. Remove the existing osds from the list of osds
3. Create the osds not already available.

The approach I used to read the json output into a list for comparison is a bit cumbersome. I somebody has a better idea feedback for a more lean version would be highly welcome.

Tested with a bare metal Proxmox 6.x cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -628,3 +628,4 @@ Michael Holasek ([@mholasek](https://github.com/mholasek))
 [acl-module]: https://github.com/lae/ansible-role-proxmox/blob/master/library/proxmox_group.py
 [storage-module]: https://github.com/lae/ansible-role-proxmox/blob/master/library/proxmox_storage.py
 [datacenter-cfg]: https://pve.proxmox.com/wiki/Manual:_datacenter.cfg
+[ceph_volume]: https://github.com/ceph/ceph-ansible/blob/master/library/ceph_volume.py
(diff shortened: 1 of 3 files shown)


lae:
@gaudenz or @trickert76 any chance one of you could test this in a physical environment?

You can install this PR using galaxy like below (and this'll let you reuse a playbook as-is):

```
ansible-galaxy install https://github.com/mholasek/ansible-role-proxmox/archive/feature/ceph-osd-idempotent.tar.gz,pr-81,lae.proxmox --force
```

If possible I'd like this to be tested on an existing cluster - making a new one using v1.6.2 and then re-running the same playbook but with this version would be the safest.

lae:
Rebased on current develop. If you need to make any changes (I'll review it more thoroughly later, but it looks fine on a cursory review), Michael, just delete your local copy of the branch and do a pull/checkout again.

trickert76:
Sorry - I'ved removed my CEPH environment and switched to GlusterFS because I can use "partitions" instead of complete harddisc. My provider cannot integrate more then 2 discs in their hardware (plus M2.. but that isn't really cheep).

lae:
Oh, forgot I made a local rebase that updated the commit hash. Anyway, I updated the Vagrantfile to provide a hard disk to use as an OSD and tested the scenario I mentioned earlier without issue.

Issue #92, opened 2020-05-16
interfaces template and resolvconf


jfpanisset:
In the section of README.md which discusses the interfaces-{{ pve_group }}.j2 template, the sample code demonstrates setting the DNS servers inside the `iface` section:

```
    dns-nameservers 10.2.2.4 10.3.2.4
    dns-search local
```

But based on [Defining the DNS Servers](https://wiki.debian.org/NetworkConfiguration#Defining_the_.28DNS.29_Nameservers), it appears that this is only the case if you have installed the `resolvconf` package: on a fresh install of Proxmox this is not present, and the original installer will have created a static `/etc/resolv.conf` file.

So perhaps this should be noted in the documentation, and/or explicit code to configure `/etc/resolv.conf` could be added, similarly to the support for configuring NTP?

lae:
That section is really just an example playbook (configuring interfaces is outside the scope of the role itself) and isn't meant to be super comprehensive because everyone's environments are different.

Thoughts on just removing those lines from the interfaces example altogether, to prevent possible confusion from those extraneous options (as far as I can tell)?

jfpanisset:
I've submittted PR #94 to suggest possible wording to discuss this issue. If I remember well the default Proxmox installer only lets you specify a single DNS server (I could be wrong), and I'm guessing that most cluster configurations would want at least 2.

trickert76:
I would remove that two lines from the example and thats it. Every environment is different. In my case I would not install resolvconf and use the /etc/resolv.conf and I have templates in a different role for setup the DNS etc.

PR #94, opened 2020-05-17
Discuss resolv.conf vs resolvconf to define DNS servers.

jfpanisset:
Based on issue #92 I am suggesting a paragraph to discuss alternative scenarios for DNS configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,6 +262,8 @@ iface enp2s0f1 inet static
     netmask 255.255.255.0
 ```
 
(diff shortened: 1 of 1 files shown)


lae:
I guess since we're doing static assignments anyway and DNS is somewhat important, I'd prefer to just bundle installation of the `resolvconf` package within the initial apt task in the given example, if only to minimize tangents in the documentation, and maybe just a small comment in the apt task beside the package name, such as `# needed for static resolver config`.

lae:
4335befdd1d570918b755715ecd1b3ccfe33d60d

Issue #23, opened 2017-10-24
Manage authentication realms in Proxmox


lae:
By default `pam` and `pve` are realms created within Proxmox for authenticating users into Proxmox. This supports most use cases. However, Proxmox supports LDAP/AD realms, which some users may want:

https://pve.proxmox.com/wiki/User_Management#pveum_authentication_realms

This role should introduce a `proxmox_realm` (or something else) module to help manage and create these realms.

For new contributors:

There are existing modules in `library/` that can be used for reference for this issue. The PVE API documentation is also available at https://pve.proxmox.com/pve-docs/api-viewer/.

trickert76:
I'd like to make a try with LDAP in the next months.

trickert76:
Hint for myself:  pvesh get /access/domains

edv-pi:
I dont know why but task Configure domains.cfg in takss/main.yml:399  is conflicting with the other domain.cfg related tasks. At some point the config gets written correct but after this task task the syntax is wrong. if i exclude this task it is working
![realms_syntax_incorrect](https://github.com/lae/ansible-role-proxmox/assets/143496892/2893a13b-7b83-4dae-9473-8c68303805d3)
![realms_syntax_correct](https://github.com/lae/ansible-role-proxmox/assets/143496892/2547e36e-c3be-4c26-ada0-e574b893b78b)

edv-pi:
removing the task from main seems to be enough to fix it

lae:
The tasks were duplicated on accident when I merged both #95 and #220 in order to keep @jfpanisset's commit as-is in history, but the second PR also had the same commit with a different hash. This could've been avoided if #220 was stacked on top of #95 (base branch), maybe, though maybe that doesn't work on PRs from forks.

This should be fixed in WIP branch `release/v1.9.0`.

lae:
Referenced PRs do introduce support for auth realms but not implemented in the manner described in this ticket, so I'm leaving this open for now as a request to implement a `proxmox_realm` or similar module. Current implementation takes full responsibility for `domain.cfg` whereas if this were implemented using a module and PVE API changes could be managed in a more off-hands approach (and we wouldn't have to risk having incorrect syntax in the file itself).

trickert76:
Oh. I forgot this ticket completely. Will have a look on it. What do you think, should be solved with the module? All realm specific getter/setter?

trickert76:
[...]

PR #95, opened 2020-05-18
Support for adding authentication realms to domains.cfg

jfpanisset:
This branch somewhat addresses issue #23 with support for adding non-default authentication domains / realms to `/etc/pve/domains.cfg` in a similar fashion to the way `/etc/pve/datacenter.cfg` is handled. This code overwrites the configuration file directly, it would be preferable if a true Python module were written to use the [domains REST API](https://pve.proxmox.com/pve-docs/api-viewer/#/access/domains/{realm}) but for now this appears to be sufficient to add Active Directory authentication to login to the web UI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,6 +415,7 @@ pve_groups: [] # List of group definitions to manage in PVE. See section on User
 pve_users: [] # List of user definitions to manage in PVE. See section on User Management.
 pve_storages: [] # List of storages to manage in PVE. See section on Storage Management.
 pve_datacenter_cfg: {} # Dictionary to configure the PVE datacenter.cfg config file.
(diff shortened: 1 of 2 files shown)


jfpanisset:
Rebased against `develop`, looking at other requests.

lae:
Thanks for rebasing. It looks like the changeset is incomplete. `pve_domains_cfg` needs to be given as a default in `defaults/main.yml`, otherwise it breaks deployment on the `Check domains.cfg exists` task:

https://travis-ci.org/github/lae/ansible-role-proxmox/jobs/690697670#L2120

lae (approved):

PR #96, opened 2020-05-24
Get rid of the Enterprise repo first

jfpanisset:
Git rid of the Enterprise repo first, or else "Add Proxmox repository" fails when it tries to
"apt update" with an "Unthauthorized IP" error for sites that don't have a Proxmox subscription.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -86,6 +86,28 @@
     name: os-prober
     state: absent
 
+- block:
+  - name: Remove automatically installed PVE Enterprise repo configuration
+    apt_repository:
+      repo: "{{ item }}"
+      filename: pve-enterprise
+      state: absent
+    with_items:
+      - "deb https://enterprise.proxmox.com/debian {{ ansible_distribution_release }} pve-enterprise"
+      - "deb https://enterprise.proxmox.com/debian/pve {{ ansible_distribution_release }} pve-enterprise"
+
+  - name: Remove subscription check wrapper function in web UI
+    patch:
+      src: "00_remove_checked_command_{{ ansible_distribution_release }}.patch"
+      basedir: /
+      strip: 1
+      backup: yes
+    when:
+      - "pve_remove_subscription_warning | bool"
+
+  when:
+    - "'pve-no-subscription' in pve_repository_line"
+
 - name: Add Proxmox repository
   apt_repository:
     repo: "{{ pve_repository_line }}"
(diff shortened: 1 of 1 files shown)


lae:
Please rebase against develop. There were a few fixes pushed to update the test suite.

The location where this block is originally placed is after PVE packages are installed, because this modifies the files that are installed themselves. I'm not sure this change works on new installations (which CI will help us see), especially when it tries to patch the javascript libraries.

Can you provide the specific steps that you are taking (role variables, what type of endpoint (clean debian installations over PXE, for example) in order to reproduce the issue this tries to solve? It hasn't been reported thus far nor has CI failed with this error previously, so I'm wondering if there's something specific to your environment.

jfpanisset:
Sorry, I think I've broken this branch (I'm not very good with rebase workflows), I'll need to clean this up first. 

I was running the role against a freshly deployed Proxmox cluster where each node had just been installed from the Proxmox 6.1-2 ISO. What I observed is that by default the ISO installer creates `/etc/apt/sources.list.d/pve-enterprise.list` so when the step `Add Proxmox repository` tried to run with that source still in place and no valid subscription, `apt` would error out with an `Unauthorized IP` and this would abort the run. 

The role variable I have in `group_vars/pve01` is:

```
pve_repository_line: "deb http://download.proxmox.com/debian/pve stretch pve-no-subscription"
```

jfpanisset:
I think I've cleaned up the branch history and it is now correctly rebased against `develop`.

lae:
Okay, I see the issue. Usage of this role isn't supported on hosts that were provisioned via the official Proxmox installation CDs. This role is (and always has been) meant to be an alternative to the official installer and is supposed to be used on generally fresh Debian installations. This is kind of implied by the galaxy meta and by the fact that this is an Ansible role for *installing* Proxmox, but after looking through the README I guess it isn't 100% clear that it's supposed to be used on bare Debian installs, not Proxmox, so I'll add something there.

CI is failing where I expected it to fail from the current set of changes: https://travis-ci.org/github/lae/ansible-role-proxmox/jobs/690696246#L1040

That particular failure can probably be resolved by splitting the block and moving the patch task after, but even then, I expect that the removal will still cause an issue on fresh Debian installations because in that case, the removal would be done before the package that creates the file is installed so on an initial deployment this change would leave the enterprise repo installed, which may trip some people up when they try to use `apt` manually on a host afterwards. This change would also break the idempotency check, I think.

There's no trivial way to set up LXC containers in the test suite from the Proxmox installation ISOs as far as I know - otherwise I could consider trying to support both scenarios (then again, it would probably double the amount of time CI takes for every test...).

If this is the only issue stopping you from using this role on the Proxmox installation ISOs and you're unable to switch to using plain Debian, I'd suggest just adding that one task in the `pre_tasks` section of your playbook.

PR #105, opened 2020-10-27
Fix trusting Proxmox' packaging key

stackcoder:
The ansible `apt_key` module requires gpg to be installed. The gpg package is not present when installing Debian without system tools.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -75,6 +75,11 @@
         "{{ hostvars[item].ansible_hostname }}"
       ]
 
+- name: Ensure gpg is installed
+  apt:
+    name: gpg
+    state: present
+
 - name: Trust Proxmox' packaging key
   apt_key:
     data: "{{ lookup('file', pve_release_key) }}"


lae:
Can you provide specific steps as to how you're bringing up Debian installations without gpg already installed? It is already part of the `standard` tasksel, which is a reasonable requirement for bringing up PVE in my opinion.

chymian:
IMHO, in the times of packer & devops, one cannot relay on ISO install with tasksel, as well as it's not relyable, to think every provider, who gives you a buster basic-install on metal, has used tasksel.
I might have missed it, since I'm started trying to use this role just recently, but neither tasksel not gpg is noted as preference.

zaneneuschuler:
I'm sorry, but considering that gpg is installed by default on literally any debian distribution, no matter what your tasksel preferences are set to, I can't see how you could possibly have gpg missing unless you intentionally remove it yourself.

stackcoder:
Installing Debian without gpg is quite easy: Just deselect everything in tasksel, even `standard system utilities`.

lae:
for my reference:
apt depends only on `gpgv` and functions fine without `gnupg`. apt-key has a soft dependency on the `gnupg` package.

Anyway, I'll merge this, but the PVE documentation states:

> Install a standard Debian Buster, for details see Debian, and configure a static IP. It is recommended to only install the "standard" package selection, and nothing else, as Proxmox VE brings its own packages for qemu, lxc.

https://pve.proxmox.com/wiki/Install_Proxmox_VE_on_Debian_Buster

lae (approved):

PR #106, opened 2020-10-27
Allow hot-reloading of network configuration via PVE UI

stackcoder:
Fix applying network configuration using the web ui.


lae (changes requested):
Per https://pve.proxmox.com/wiki/Network_Configuration this should be optional and not forced on the user in case the administrator has issues with ifupdown2.

PR #110, opened 2020-12-08
Support management of user roles

trickert76:
I've added a new module for managing pve roles (because Icinga monitoring user needs a special role and I don't want to add it manually on all clusters). I was able to create and remove a role.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -1,5 +1,5 @@
----
 # defaults file for ansible-role-proxmox
+---
 pve_group: proxmox
 pve_fetch_directory: fetch
 pve_repository_line: "deb http://download.proxmox.com/debian/pve {{ ansible_distribution_release }} pve-no-subscription"
@@ -34,9 +34,14 @@ pve_cluster_clustername: "{{ pve_group }}"
 pve_datacenter_cfg: {}
 pve_cluster_ha_groups: []
 pve_ssl_letsencrypt: false
+# additional roles for your cluster (f.e. for monitoring)
+pve_roles: []
+# additional groups
 pve_groups: []
+# additional users (except root)
 pve_users: []
 pve_acls: []
 pve_storages: []
+# SSH port for communication between cluster nodes
 pve_ssh_port: 22
 pve_manage_ssh: true
--- a/library/proxmox_role.py
+++ b/library/proxmox_role.py
@@ -1,0 +1,192 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+ANSIBLE_METADATA = {
+    'metadata_version': '0.1',
+    'status': ['preview'],
+    'supported_by': 'lae'
+}
+
+DOCUMENTATION = '''
+---
+module: proxmox_role
+
+short_description: Manages the Access Control List in Proxmox
+
+options:
+    name:
+        required: true
+        description:
+            - name of the role.
+    privileges:
+        required: true
+        type: list
+        description:
+            - Specifies a list of PVE privileges for the given role.
+    state:
+        required: false
+        default: "present"
+        choices: [ "present", "absent" ]
+        description:
+            - Specifies whether this role should exist or not.
+
+author:
+    - Thoralf Rickert-Wendt (@trickert76)
+'''
+
+EXAMPLES = '''
+- name: Create a role for monitoring with given privileges
+  proxmox_role:
+    name: "monitoring"
+    privileges: [ "Sys.Modify", "Sys.Audit", "Datastore.Audit", "VM.Monitor", "VM.Audit" ]
+'''
+
+RETURN = '''
+'''
+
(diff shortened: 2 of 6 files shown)


library/proxmox_role.py
```diff
@@ -0,0 +1,178 @@
...
+            return (False, e.message)
+
+    def modify_role(self):
+        existing_role = self.lookup()
+        modified_role = self.prepare_role_args()
+        updated_fields = []
+        error = None
+
+        for key in modified_role:
+          updated_fields.append(key)
```
lae, line 117: I think the way this is written always results in updated fields, given `prepare_role_args()` parses the task values every time? Is there supposed to be a comparison to `existing_role` here?

library/proxmox_role.py
```diff
@@ -0,0 +1,178 @@
...
+            # No changes necessary
+            return (updated_fields, error)
+
+        try:
+            pvesh.set("access/roles/{}".format(self.name), **modified_role)
+        except ProxmoxShellError as e:
+            error = e.message
+
+        return (updated_fields, error)
+def main():
```
lae, line 132: need newline here

defaults/main.yml
```diff
@@ -1,5 +1,5 @@
----
 # defaults file for ansible-role-proxmox
+---
```
lae, line 2: not sure if this might irritate the linter. if it doesn't complain, then it's fine.

tasks/main.yml
```diff
@@ -185,13 +185,21 @@
...
+  when: "not pve_cluster_enabled | bool or (pve_cluster_enabled | bool and inventory_hostname == groups[pve_group][0])"
+
 - name: Configure Proxmox groups
   proxmox_group:
     name: "{{ item.name }}"
     state: "{{ item.state | default('present') }}"
     comment: "{{ item.comment | default(omit) }}"
   with_items: "{{ pve_groups }}"
-  when: "not pve_cluster_enabled or (pve_cluster_enabled and inventory_hostname == groups[pve_group][0])"
+  when: "not pve_cluster_enabled | bool or (pve_cluster_enabled | bool and inventory_hostname == groups[pve_group][0])"
```
lae, line 202: These changes should preferably be done in a separate PR as they're not associated with this feature.

lae (changes requested):
Overall looks good to me.

When you're making changes after reviewing my comments, can you squash all your commits into one commit and force push that?

defaults/main.yml
lae: just checked, looks like it doesn't

trickert76: actually - this should be "normal". Everything above --- is a header and must be commented. It is designed in YAML to contain a description, author ets. My personal project with ansible roles runs agains ansible-lint and yamllint and that is ok.

tasks/main.yml
trickert76: ok

library/proxmox_role.py
trickert76: Yes, thats right. I'm still working on it.

library/proxmox_role.py
trickert76: done

trickert76:
> Overall looks good to me.
> 
> When you're making changes after reviewing my comments, can you squash all your commits into one commit and force push that?

Can you give me a hint, how to do that? I'm finished now. The proxmox_role works.

lae:
Have a read through the following article:

https://gitbetter.substack.com/p/how-to-squash-git-commits

You can run `git rebase -i develop` for an easy way to start off from the develop branch. Leave the first commit as pick, change the rest to squash. You should remove any of the extraneous commit messages when you're editing the final commit message so that it's succinct and describes the end result (e.g. the final commit history won't have the eclipse settings file, so there shouldn't be a mention of it).

Once you've squashed those commits into one, use `git push -f` to overwrite the remote branch (on github) and this PR will reflect that change automatically.

library/proxmox_role.py
```diff
@@ -109,12 +111,23 @@ def create_role(self):
...
 
         for key in modified_role:
-          updated_fields.append(key)
+            if key not in existing_role:
+                updated_fields.append(key)
+            else:
+                new_value = modified_role.get(key)
+                old_value = existing_role.get(key)
+                if isinstance(old_value, list):
+                    old_value = str(old_value.sort())
```
lae, line 125: `sort()` doesn't return the contents of the list here. You'd need to sort it separately and then use `old_value`/`new_value`.  (I'm also not sure why `str()` is being used here instead of like, `join()`.)

trickert76: Ah, you're right. That code is irrelevant. The value is always a string.

trickert76:
I'm to stupid for squashing - i tried different ways but at the end there are always just new commits to origin. So I'll open a new PR with a fresh branch

PR #111, opened 2020-12-10
Yaml-Syntax Fixes (Indents, Bool, Reboot-Module)

trickert76:
Made some minor fixes for better YAML syntax

- use | bool filter for unknown var types
- use quotes in when conditions
- correct indentions within blocks
- replace shell script for reboot with existing reboot-module (since Ansible 2.6 available)


zenntrix (approved):
Thanks for taking the time to making these improvements, they all look reasonable :-)

PR #114, opened 2021-01-01
Add ZFS Storage support

foerkede:
I added support for the zfspool storage type. I also added the possibility to use all content types on a zfs volume, let me know if my solution is ok like this.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -226,6 +226,14 @@
   with_items: "{{ pve_acls }}"
   when: "not pve_cluster_enabled | bool or (pve_cluster_enabled and inventory_hostname == groups[pve_group][0])"
 
+- name: Create ZFS Pools
+  zfs:
+    name: "{{ item.pool }}"
+    state: present
+  with_items: "{{ pve_storages }}"
+  when: item.type == "zfspool"
+  tags: storage
+
 - name: Configure Proxmox Storage
   proxmox_storage:
     name: "{{ item.name }}"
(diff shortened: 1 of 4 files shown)


zenntrix (commented):
It looks good from a visual inspection, can you confirm what happens if you re-run the playbook on an existing cluster with the ZFS pools present?

I could test it myself but don't currently have a sacrificial cluster to play with :-)

tasks/main.yml
```diff
@@ -226,6 +226,21 @@
   with_items: "{{ pve_acls }}"
   when: "not pve_cluster_enabled | bool or (pve_cluster_enabled and inventory_hostname == groups[pve_group][0])"
 
+- name: Create ZFS Pools
+  zfs:
+    name: "{{ item.pool }}"
+    state: present
+  with_items: "{{ pve_storages }}"
+  when: item.type == "zfspool"
```
lae, line 234: wrap entire conditional with quotes

lae (changes requested):
Overall looks good to me. Can't imagine there really being an issue with existing zpools, though it'd be nice to have someone test that out (I might be able to do it in my homelab later but I'd have to do some backup prep).

I feel like this feature can be covered by automated testing. Could you try to implement tests for this underneath `tests/group_vars`? There's already a `pve_storages` query check in `tests/test.yml` so an extra validation check probably won't be necessary. Any setup work outside of the role goes in `tests/install.yml` (though it doesn't seem like this would be necessary).

Also, please rebase against develop and force push your changes here. (`git pull` on develop branch and `git rebase -i develop` on your feature branch)

foerkede:
> can you confirm what happens if you re-run the playbook on an existing cluster with the ZFS pools present?

Since I'm using the ansible zfs module there should be no problems, if it exists it does nothing and works as expected in my tests.

foerkede:
I finally did it, I kinda forgot about it...
I hope everything is right, it's my first time contributing to a public repo :)

lae (approved):

lae:
https://travis-ci.org/github/lae/ansible-role-proxmox/jobs/754730245#L2257

I forgot about loading a kernel module for zfs. It's intentionally disabled for LXC since of isolation issues so I'm leaving those commits out. May add testing for them in the future when I moved to self-hosted CI.

PR #115, opened 2021-01-19
change acl "propagate" attribute to integer

phemmer:
The `get access/acl` call returns `propagate` as an integer value. However the module was defining it as a string. This resulted in the module constantly redefining the ACL and reporting it as "changed" when nothing was actually changing.

Not sure if at some point in the past this actually was a string, but as of 6.3 it is an integer.

### DIFF
--- a/library/proxmox_acl.py
+++ b/library/proxmox_acl.py
@@ -99,7 +99,7 @@ class ProxmoxACL(object):
             for constituent in constituents:
                 self.acls.append({
                     "path": self.path,
-                    "propagate": "1", # possibly make this configurable in the module later
+                    "propagate": 1, # possibly make this configurable in the module later
                     "roleid": role,
                     "type": constituent[0],
                     "ugid": constituent[1]


zenntrix (approved):

lae:
This change seems to break the idempotence test in PVE 5 (and I'm not sure/haven't looked into why the idempotence test failed for PVE 6 in the storage module right after):
https://travis-ci.org/github/lae/ansible-role-proxmox/jobs/755217477#L2243

Idempotence tests passed for both PVE 5 and 6 in the run 4 days ago:
https://travis-ci.org/github/lae/ansible-role-proxmox/jobs/754729518

Can you provide some role variable definitions you were using that caused you to run into this issue?

phemmer:
Sure, I still have output from earlier debugging:

Here's where I added `result['expected'] = acl.acls` and `result['actual'] = acl.existing_acl` to `main()`. `item` shows the input definition.
```
changed: [myhost.mydomain] => (item={'path': '/', 'groups': ['admin'], 'roles': ['Administrator']}) => {
    "actual": [
        {
            "path": "/",
            "propagate": 1,
            "roleid": "Administrator",
            "type": "group",
            "ugid": "admin"
        }
    ],
    "ansible_loop_var": "item",
    "changed": true,
    "expected": [
        {
            "path": "/",
            "propagate": "1",
            "roleid": "Administrator",
            "type": "group",
            "ugid": "admin"
        }
    ],
    "item": {
        "groups": [
            "admin"
        ],
        "path": "/",
        "roles": [
            "Administrator"
        ]
    },
    "state": "present"
}
```

lae:
Added a bit.

https://github.com/lae/ansible-role-proxmox/commit/8e69cb6b7e27b05321bea28d3afa32a2fb8b47ce
https://travis-ci.org/github/lae/ansible-role-proxmox/builds/755283588

PR #121, opened 2021-03-08
Add autoscale mode for Ceph pool configuration

slobberbone:
Ability to manage autoscale mode for pool's Ceph configuration

This Ceph pool uses custom autoscale mode : "off" | "on" | "warn"> (default = "warn")
```
  - name: vm-storage,
    pgs: 128,
    rule: replicated_rule,
    application: rbd,
    autoscale_mode: "on",
    storage: true,
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -649,6 +649,13 @@ pve_ceph_pools:
     storage: true
     size: 2
     min-size: 1
+# This Ceph pool uses custom autoscale mode : "off" | "on" | "warn"> (default = "warn")
+  - name: vm-storage,
+    pgs: 128,
+    rule: replicated_rule,
+    application: rbd,
+    autoscale_mode: "on",
+    storage: true,
 pve_ceph_fs:
 # A CephFS filesystem not defined as a Proxmox storage
   - name: backup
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -649,6 +649,13 @@ pve_ceph_pools:
     storage: true
     size: 2
     min-size: 1
+# This Ceph pool uses custom autoscale mode : "off" | "on" | "warn"> (default = "warn")
+  - name: vm-storage,
```
lae, line 653: YAML doesn't expect commas here.

lae (changes requested):

README.md
slobberbone: arg ... on my side I've in my yaml something like : ```pve_ceph_pools: [
        {
          name: vm-storage,
          pgs: "{{ ceph_pool_pgs }}",
          rule: replicated_rule,
          application: rbd,
          autoscale_mode: "on", #"off" | "on" | "warn"> (default = "warn")
          storage: true,
        }
        ],```
that's why ... I'll fix that

lae (approved):

PR #122, opened 2021-03-08
Add group for deploying multiple Ceph managers

slobberbone:
Ability to configure Ceph nodes as manager

### DIFF
--- a/tasks/ceph.yml
+++ b/tasks/ceph.yml
@@ -45,6 +45,12 @@
     creates: '/var/lib/ceph/mon/ceph-{{ ansible_hostname }}/'
   when: "inventory_hostname != groups[pve_ceph_mon_group][0]"
 
+- name: Create additional Ceph managers
+  command: 'pveceph mgr create'
+  args:
+    creates: '/var/lib/ceph/mgr/ceph-{{ ansible_hostname }}/'
+  when: "inventory_hostname in groups[pve_ceph_mgr_group]"
+
 - block:
     - name: Get existing ceph volumes
       ceph_volume:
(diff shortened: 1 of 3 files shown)


lae (commented):
Setting `pve_ceph_mgr_group` to `pve_group` default makes it so that ceph-mgr is deployed on all hosts, which I don't think is default behaviour? This may be an unwanted change for existing deployments.

Anyone else have any thoughts on this?

slobberbone:
For now, manager is only on one node ... if you want to have a manager on each, it's not possible. That's why this PR ;) If you prefer to set it only on one by default, tell me

slobberbone:
I can set `pve_ceph_mgr_group: "{{ groups[pve_group][0] }}"` in the default/main.yml if you want, like this, it's not changing anything by default

slobberbone:
In fact I had some warning on Ceph with som delay between time on servers, it was solve configurin NTP on each nodes + enable manager on each nodes : https://forum.proxmox.com/threads/ceph-osd-on-pve5-got-timeout-500-in-gui.36235/
Another point is that if your manager going down, you don't have any manager at all ...
https://pve.proxmox.com/wiki/Deploy_Hyper-Converged_Ceph_Cluster#pve_ceph_manager
If you set a manager on 32 nodes, only one is running at one time, so for me it makes sense to have one (like before) by default to not change the behaviour of the project, and the capabilty to run it on a group (may be all)

lae:
> it was solved configuring NTP on each nodes + enable manager on each nodes 

I think the manager thing is actually unrelated here, and the instructions in this role already say to configure NTP.

> https://pve.proxmox.com/wiki/Deploy_Hyper-Converged_Ceph_Cluster#pve_ceph_manager

It's been a while since I used Ceph, but the doc here reminded me that the recommended configuration is to have it set the same to the group of Ceph monitors. So I guess I'd be fine with the default being set to the same group as the monitor group.

And on another note, I was also kind of thinking this role's default behaviour for Ceph monitors is also a bit overkill, too, so it might be a good idea to change that and have it be a required variable if Ceph is enabled. This would be a separate PR though.

slobberbone:
So you prefer to use monitor group instead of manager ... it was just to have the possibility to do what you want ...

lae:
That's not what I said at all. I'm talking about the **default** being set to the monitor group, where right now it is set to the cluster group.

tasks/ceph.yml
```diff
@@ -45,6 +45,12 @@
     creates: '/var/lib/ceph/mon/ceph-{{ ansible_hostname }}/'
   when: "inventory_hostname != groups[pve_ceph_mon_group][0]"
 
+- name: Create additional Ceph managers
+  command: 'pveceph mgr create'
+  args:
+    creates: '/var/lib/ceph/mgr/ceph-{{ ansible_hostname }}/'
+  when: inventory_hostname in groups[pve_ceph_mgr_group]
```
lae, line 52: This should be in quotes.

lae (approved):

PR #123, opened 2021-03-08
Correct list of required attrs for RBD storages

slobberbone:
When you configure a pool with rdb, set username and path is not required.


lae (approved):

lae:
https://pve.proxmox.com/pve-docs/chapter-pvesm.html#ceph_rados_block_devices

PR #124, opened 2021-03-08
Add the ability to disable hosts file management

slobberbone:
I've configured several servers at different providers to be linked together throw VPN (i'm using Tinc) I that case, my hosts file have to be configured before launching proxmox tasks.
By default, it's change nothing compare to before. If you set the variable to "no", then tasks configuring hosts file are skipped

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -37,7 +37,7 @@
 
 
       {% endfor %}"
-  when: "pve_cluster_enabled | bool"
+  when: "pve_manage_hosts_enabled | bool"
 
 - name: Remove conflicting lines in hosts files
   lineinfile:
@@ -76,7 +76,7 @@
         "{{ hostvars[item].ansible_fqdn }}",
         "{{ hostvars[item].ansible_hostname }}"
       ]
-  when: "pve_cluster_enabled | bool"
+  when: "pve_manage_hosts_enabled | bool"
 
 - name: Ensure gpg is installed
   apt:
@@ -135,6 +135,7 @@
   apt:
     name: "{{ _pve_install_packages }}"
     state: "{{ 'latest' if pve_run_proxmox_upgrades else 'present' }}"
+    update_cache: yes
   retries: 2
   register: _proxmox_install
   until: _proxmox_install is succeeded
@@ -155,6 +156,7 @@
         basedir: /
         strip: 1
         backup: yes
+      ignore_errors: yes
       when:
         - "pve_remove_subscription_warning | bool"
   when:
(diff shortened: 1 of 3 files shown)


tasks/main.yml
```diff
@@ -135,6 +135,7 @@
   apt:
     name: "{{ _pve_install_packages }}"
     state: "{{ 'latest' if pve_run_proxmox_upgrades else 'present' }}"
+    update_cache: yes
```
lae, line 138: This should be a separate PR. `cache_valid_time` should also be set so it doesn't trigger idempotency failures during tests.

tasks/main.yml
```diff
@@ -155,6 +156,7 @@
         basedir: /
         strip: 1
         backup: yes
+      ignore_errors: yes
```
lae, line 159: We don't want to ignore errors here; we want to catch failures from this during weekly testing.

If a user runs into an error with this task and they don't want to fix it, they can just set `pve_remove_subscription_warning` to false.

Not to mention this isn't really related to this PR, anyway.

tasks/main.yml
```diff
@@ -37,7 +37,7 @@
 
 
       {% endfor %}"
-  when: "pve_cluster_enabled | bool"
+  when: "pve_manage_hosts_enabled | bool"
```
lae, line 40: This conditional should not replace the check for clustering. Both should be checked.

tasks/main.yml
```diff
@@ -76,7 +76,7 @@
         "{{ hostvars[item].ansible_fqdn }}",
         "{{ hostvars[item].ansible_hostname }}"
       ]
-  when: "pve_cluster_enabled | bool"
+  when: "pve_manage_hosts_enabled | bool"
```
lae, line 79: should not be replaced

lae (changes requested):
Please amend your commit to remove the extraneous stuff and fix the conditionals (so that when I merge it, it'll be clean/the extraneous changes won't exist).

```
git checkout configuration-hosts-file-skip
# edit your files here
git commit --amend
git push --force
```

tasks/main.yml
slobberbone: oups ... sorry I don't see that, it just because it was in error when I execute it and for me, this is not a big deal if it's wrong ... but get it. I remove it !

tasks/main.yml
slobberbone: Ok

tasks/main.yml
slobberbone: done

tasks/main.yml
slobberbone: done too

lae (approved):

PR #130, opened 2021-04-28
Add additional ceph-mon when host in mon group

PendaGTP:
Install Ceph Monitor daemon only on hosts of the group `pve_ceph_mon_group`


zenntrix (approved):
Seems like a sensible and reasonable fix :-)

PR #131, opened 2021-04-28
Add ability to specify on wich nodes Ceph is installed

PendaGTP:
Ability to choose on which nodes Ceph is installed.
This is very useful when you want to separate the 'compute nodes' and the 'storage nodes' on Proxmox cluster.

This RP is related to: #130


PendaGTP:
This PR allows to realize this kind of architecture, the goal is to separate the compute part and the storage part.
proxmox-ceph-cluster

zenntrix (approved):

Issue #34, opened 2018-05-25
Support in-PVE Let's Encrypt setup


lae:
PVE 5.2 introduced ACME support within PVE.

https://pve.proxmox.com/wiki/Certificate_Management

An ACME management module can be created to manage ACME registration and certificate.

Process I used to generate cert:

```
~# pvesh create /cluster/acme/account -contact $email -tos_url https://letsencrypt.org/documents/LE-SA-v1.2-November-15-2017.pdf 
Generating ACME account key..
Registering ACME account..
Registration successful, account URL: 'https://acme-v02.api.letsencrypt.org/acme/acct/X'
200 OK
~# pvesh set /nodes/maika/config -acme domains=$fqdn
200 OK
~# pvesh create /nodes/maika/certificates/acme/certificate -force
Loading ACME account details
Placing ACME order
Order URL: https://acme-v02.api.letsencrypt.org/acme/order/X/X

Getting authorization details from 'https://acme-v02.api.letsencrypt.org/acme/authz/XXXXXXXXXXXXX'
... pending!
Setting up webserver
Triggering validation
Sleeping for 5 seconds
Status is 'valid'!

All domains validated!

Creating CSR
Finalizing order
Checking order status
valid!

Downloading certificate
Setting pveproxy certificate and key
Restarting pveproxy
200 OK
UPID:XXX
```

Note that failure returns 200 OK:
```
Triggering validation
Sleeping for 5 seconds
Status is still 'pending', trying again in 30 seconds
validating challenge 'https://acme-v02.api.letsencrypt.org/acme/authz/XXXXXXXX' failed
200 OK
```
Documentation should note that port 80 must be accessible from the Internet (above happened to me because firewall dropped connections) and nothing else should be running on port 80.

lae:
For the record, I think I had this working in the 'feature/letsencrypt` branch back in September but haven't merged it yet because of distractions.

trickert76:
Have a look at that

```
- name: Install Proxmox ACME Dependencies (Python2)
  apt: 
    name: ['python-pexpect']
    state: present
    update_cache: yes

- name: Install Proxmox ACME Dependencies (Python3)
  apt: 
    name: ['python3-pexpect']
    state: present
    update_cache: yes

- name: Verify ACME account
  command: pvenode acme account list
  register: __acme_account

- name: Create ACME Account 
  expect:
[...]

PR #147, opened 2021-08-04
Feature/letsencrypt

trickert76:
see #34


virtualized-human:
is there something missing in the implementation since this seems to be sadly dead?

PR #148, opened 2021-08-07
Use linkN args for cluster init also on bullseye 

mweinelt:
In the same vein as https://github.com/lae/ansible-role-proxmox/commit/2121826548a6b009edf3b23e03512be3fd435f68.


lae (changes requested):
I haven't reviewed Richard's changes in the feature branch yet (assuming not ready for review since no PR) but since this branch should be removing support for PVE 5/Stretch, we shouldn't have to test against distro version anymore and move the linkN flags into the `command` value itself.

PR #149, opened 2021-08-07
Switch managed ssh keys to ed25519

mweinelt:
- Migrates the generated and managed keys to ed25519.
- Manages root authorized_keys exclusively
- Allows additional keys via `pve_ssh_extra_keys`

### DIFF
--- a/tasks/ssh_cluster_config.yml
+++ b/tasks/ssh_cluster_config.yml
@@ -9,19 +9,20 @@
   user:
     name: root
     generate_ssh_key: yes
-    ssh_key_file: /root/.ssh/id_rsa
-    ssh_key_type: rsa
+    ssh_key_bits: 521
+    ssh_key_file: /root/.ssh/id_ed25519
+    ssh_key_type: ed25519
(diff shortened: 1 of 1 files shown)


lae:
Could we get some testing done to see how this affects existing clusters? (i.e. existing cluster using RSA keys, then a playbook with these changes get applied, then an attempt to add a new node with this playbook)

While I'd prefer the switch to ed25519, upstream hasn't done so, so I'm not sure how this affects support by official channels, too. But the conversation at https://bugzilla.proxmox.com/show_bug.cgi?id=1954 seems to suggest that they might (and that there might be more necessary changes here, not sure).

As for `pve_ssh_extra_keys`, I think this may get in the way of existing installations that are managing other SSH keys differently (like maybe outside of Ansible)?

virtualized-human:
im looking to get exact that what is implemented with `pve_ssh_extra_keys`. I dont think that this will get in the way? Why should it be if the default is not set and so no one is required to use that feature?

virtualized-human:
anyways will test it now with reeimplementing in current version

lae (approved):

PR #150, opened 2021-08-07
Don't run corosync on default IPv4 network by default

mweinelt:
This is error prone, if the host does not have IPv4 connectivity, and
worse this would run the corosync network on a public network, which
isn't an overall great design.

I believe it is more reasonable to have users of the role explicitly
decide on their cluster network.

---

In the first case `ansible_default_ipv4` is an empty attribute set and the task fails when accessing `.address` on it.

This strictly improves compatibility with IPv6 only hypervisors.

### DIFF
--- a/vars/main.yml
+++ b/vars/main.yml
@@ -2,6 +2,3 @@
 # vars file for ansible-role-proxmox
 pve_base_dir: "/etc/pve"
 pve_cluster_conf: "{{ pve_base_dir }}/corosync.conf"
-
-# defaults that need to be host facts
-_pve_cluster_addr0: "{{ ansible_default_ipv4.address }}"
--- a/tasks/load_variables.yml
+++ b/tasks/load_variables.yml
@@ -31,10 +31,6 @@
       when: "pve_cluster_link1_addr is defined and ansible_distribution_release == 'buster'"
   when: "pve_cluster_addr1 is not defined"
 
-- name: Define pve_cluster_addr0 if not provided
-  set_fact:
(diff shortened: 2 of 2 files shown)


vars/main.yml
```diff
@@ -2,6 +2,3 @@
 # vars file for ansible-role-proxmox
 pve_base_dir: "/etc/pve"
 pve_cluster_conf: "{{ pve_base_dir }}/corosync.conf"
-
-# defaults that need to be host facts
-_pve_cluster_addr0: "{{ ansible_default_ipv4.address }}"
```
mweinelt, line 7: Maybe like this:

```yaml
# defaults that need to be host facts
_pve_cluster_addr0: "{{ lookup('subelements', 'ansible_default_ipv4', 'address') | default(lookup('subelements', 'ansible_default_ipv6', 'address')) }}"
```

lae: Doesn't look like that'll work:

> An unhandled exception occurred while templating '{{ lookup('subelements', ansible_default_ipv4, 'address') | default(lookup('subelements', ansible_default_ipv6, 'address')) }}'. Error was a <class 'ansible.errors.AnsibleError'>, original message: An unhandled exception occurred while running the lookup plugin 'subelements'. Error was a <class 'ansible.errors.AnsibleError'>, original message: subelements lookup expects a dictionary, got '10.x.x.x'

Seems like the subelements lookup expects dicts for all its arguments (and returns a dict):

> Subelements walks a list of hashes (aka dictionaries) and then traverses a list with a given (nested sub-)key inside of those records.

lae:
Superseded by #197/#200

PR #151, opened 2021-08-07
Don't manage SSH ciphers on bullseye 

mweinelt:
The OpenSSH daemon comes with reasonable defaults these days, so don't
try to manage that, by copying the same cipher list from release to
release.

Between Debian releases there is usually a large overlap over compatible
ciphers, so there shouldn't be any fear of breaking compatibility
between Proxmox VE releases either.

---

It might be reasonable to do the same on buster, but since I'm starting
a 7.0 cluster right now and I don't know the entire reasoning behind this
setting, I'm erring on the conservative side.

### DIFF
--- a/tasks/ssh_cluster_config.yml
+++ b/tasks/ssh_cluster_config.yml
@@ -85,3 +85,13 @@
     insertbefore: BOF
     create: yes
     dest: /root/.ssh/config
+    state: present
+  when: pve_ssh_ciphers is defined
+
+- name: Drop PVE-provided ciphers, when undefined
+  lineinfile:
+    regexp: "^Ciphers .*"
+    dest: /root/.ssh/config
+    state: absent
+  when: pve_ssh_ciphers is not defined
+
--- a/vars/debian-bullseye.yml
+++ b/vars/debian-bullseye.yml
@@ -1,5 +1,4 @@
 ---
 pve_release_key: proxmox-ve-release-7.x.asc
 pve_release_key_id: DD4BA3917E23BF59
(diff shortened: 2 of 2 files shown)


lae:
`pvecm add` (when you go to manually stand up a cluster) iirc does this to disable weak ciphers, so this task emulates that behaviour.

mweinelt:
No, it is not erroring out, but it is also just pinning the current (openssh 7.9) default cipher list and therefore provides no added value. 

From the `man 5 sshd_config`
```
The default is:

chacha20-poly1305@openssh.com,
aes128-ctr,aes192-ctr,aes256-ctr,
aes128-gcm@openssh.com,aes256-gcm@openssh.com
```



Ideally one would manage acceptable ciphers, as well as MACs and KexAlgorithms in an sshd_config using Mozillas recommendations, but that is probably out of scope for this role.

https://infosec.mozilla.org/guidelines/openssh

lae:
I guess there was another reason - from: https://git.proxmox.com/?p=pve-cluster.git;a=blob;f=data/PVE/Cluster/Setup.pm;h=1f064775af3cc4b2ac12a59d7d8abb7283f4b998;hb=HEAD#l162

```
# changed order to put AES before Chacha20 (most hardware has AESNI)
```

I'm not sure if making something that's hardcoded in PVE itself optional has any effect on support (this role is meant to mimic PVE tooling wherever it can't use them directly), but this seems like a relatively harmless thing to change in the first place. I'll wait to see if anyone else has any thoughts.

mweinelt:
Pretty sure that we should prefer chacha20-poly1305 these days, in spite of AES-NI support in hardware, I trust the OpenSSH upstream alot more to make good decisions.

I feel we as sysadmins are in a good position to make better choices that don't break any compatibility.

lae:
Rebased against develop to at least cleanup the unrelated/unmerged commits from the old pve7 branch.

PR #164, opened 2021-10-05
fix detection for encrypted ceph volumes

wu3rstle:
With the encrypted flag set it is not possible to skip volume creation for existing volumes via diff.


lae:
Suggested changeset isn't a proper way to fix detection as it duplicates every entry in the comparison whether it is encrypted or not. It's also not comprehensive as it'll miss combinations of `block.db` and `block.wal`.

This PR has been superseded by #171.

PR #178, opened 2021-11-23
added usage of official ceph volume plugin

wu3rstle:
To prevent compatibility issues between roles in the same playbook the official ceph_volume module from https://github.com/ceph/ceph-ansible should be used in this role.

### DIFF
--- a/library/ceph_volume.py
+++ b/library/ceph_volume.py
@@ -2,10 +2,17 @@
 
 from ansible.module_utils.basic import AnsibleModule
 try:
-    from ansible.module_utils.ca_common import exec_command
+    from ansible.module_utils.ca_common import exec_command, \
+                                               is_containerized, \
+                                               fatal
 except ImportError:
-    from module_utils.ca_common import exec_command
+    from module_utils.ca_common import exec_command, \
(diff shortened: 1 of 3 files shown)


lae:
Alternatively, we can just rename the file and the one task (or two?) that use it? If you're using this role and the ceph-ansible roles in the same playbook this just temporarily resolves the issue of diverging modules. We were definitely out of date for over a year, anyway.

wu3rstle:
Renaming the module and module_utils name sounds ok.

I prefer prefixing the files with `pve_`  so the used module will be `pve_ceph_volume` with the imported module util `pve_ca_common`.

@lae what do yout think?

lae:
@wu3rstle yeah, prefixing the name with `pve_` sounds fine.

Also, I think the main reason why `ca_common.py` exists at all was because those functions were being used in multiple ceph ansible modules in the ceph-ansible project, but we're only using one function in it (got rid of the container-related functions since we don't interact with containerized ceph) and one module. I don't think it's very likely we'd import any other module, so maybe we can just remove that `ca_common.py` file entirely and move the function definition into the `pve_ceph_volume` module?

PR #200, opened 2022-08-11
Clustering role variable cleanup

lae:
Replaces #197


lae:
vagrant deploy passed

PR #202, opened 2022-08-15
feature/vfio-support

ol3d:
Added basic support for PCIe Passthrough using Intel or AMD based CPU's.
Feature is disabled by default as to prevent breaking change. Mediated
devices are supported, however, this is a modifiable attribute since not
all devices support gtv-g. IOMMU unsafe interupts are not necessary but
can be beneficial so support was added for that as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,7 +395,7 @@ pve_zfs_enabled: no # Specifies whether or not to install and configure ZFS pack
 pve_zfs_create_volumes: [] # List of ZFS Volumes to create (to use as PVE Storages). See section on Storage Management.
 pve_ceph_enabled: false # Specifies wheter or not to install and configure Ceph packages. See below for an example configuration.
 pve_ceph_repository_line: "deb http://download.proxmox.com/debian/ceph-pacific bullseye main" # apt-repository configuration. Will be automatically set for 6.x and 7.x (Further information: https://pve.proxmox.com/wiki/Package_Repositories)
-pve_ceph_network: "{{ (ansible_default_ipv4.network +'/'+ ansible_default_ipv4.netmask) | ipaddr('net') }}" # Ceph public network
+pve_ceph_network: "{{ (ansible_default_ipv4.network +'/'+ ansible_default_ipv4.netmask) | ansible.utils.ipaddr('net') }}" # Ceph public network
 # pve_ceph_cluster_network: "" # Optional, if the ceph cluster network is different from the public network (see https://pve.proxmox.com/pve-docs/chapter-pveceph.html#pve_ceph_install_wizard)
 pve_ceph_nodes: "{{ pve_group }}" # Host group containing all Ceph nodes
 pve_ceph_mon_group: "{{ pve_group }}" # Host group containing all Ceph monitor hosts
@@ -709,8 +709,9 @@ pve_ceph_fs:
     mountpoint: /srv/proxmox/backup
 ```
 
-`pve_ceph_network` by default uses the `ipaddr` filter, which requires the
-`netaddr` library to be installed and usable by your Ansible controller.
+`pve_ceph_network` by default uses the `ansible.utils.ipaddr` filter, which
+requires the `netaddr` library to be installed and usable by your Ansible
+controller.
 
 `pve_ceph_nodes` by default uses `pve_group`, this parameter allows to specify
 on which nodes install Ceph (e.g. if you don't want to install Ceph on all your
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -19,7 +19,7 @@ pve_zfs_enabled: no
 pve_zfs_create_volumes: []
 pve_ceph_enabled: false
 pve_ceph_repository_line: "deb http://download.proxmox.com/debian/{% if ansible_distribution_release == 'buster' %}ceph-nautilus buster{% else %}ceph-quincy bullseye{% endif %} main"
-pve_ceph_network: "{{ (ansible_default_ipv4.network +'/'+ ansible_default_ipv4.netmask) | ipaddr('net') }}"
+pve_ceph_network: "{{ (ansible_default_ipv4.network +'/'+ ansible_default_ipv4.netmask) | ansible.utils.ipaddr('net') }}"
 pve_ceph_nodes: "{{ pve_group }}"
 pve_ceph_mon_group: "{{ pve_group }}"
 pve_ceph_mgr_group: "{{ pve_ceph_mon_group }}"
--- a/tasks/ceph.yml
+++ b/tasks/ceph.yml
@@ -241,7 +241,7 @@
     path: '{{ item.mountpoint }}'
     src: |-
       {% for h in groups[pve_ceph_mon_group] -%}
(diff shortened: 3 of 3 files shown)


lae:
I see you've updated the role variable names, cool. Asides from that, can you enable this for the tests in `tests/`? Mainly need it for `tests/vagrant/` as I haven't gotten around to fixing/re-enabling CI. I'll do a further review when I can.

deloa6133:
As this is the first time I have used Vagrant, I messaged on Discord for help if you don't mind :)

tasks/pcie-passthrough.yml
```diff
@@ -0,0 +1,41 @@
...
+- block:
+  - name: Enable CPU IOMMU in default grub
+    ansible.builtin.lineinfile:
+      path: /etc/default/grub
+      regexp: ^GRUB_CMDLINE_LINUX_DEFAULT=
+      line: GRUB_CMDLINE_LINUX_DEFAULT="quiet iommu=pt{% if '"GenuineIntel" in ansible_processor | unique' %} intel_iommu=on{% else %} amd_iommu=on{% endif %}{% if pve_mediated_devices_enabled %} i915.enable_gvt=1{% endif %}"
+    register: iommu_modify
+  - name: Update default grub
+    ansible.builtin.command: update-grub
+    when: iommu_modify.changed
```
lae, line 11: Should be "iommu_modify is changed" to pass lint checker.

tasks/pcie-passthrough.yml
```diff
@@ -0,0 +1,41 @@
+---
```
lae, line 1: This file was named `pcie-passthrough.yml` but `pcie_passthrough.yml` is what is being included in `tasks/main.yml`. This file should be renamed with an underscore instead of hyphen.

tasks/pcie-passthrough.yml
```diff
@@ -0,0 +1,41 @@
...
+- block:
+  - name: Enable CPU IOMMU in default grub
```
lae, line 3: Tasks need to be indented by four spaces in a block, not two.

tasks/pcie-passthrough.yml
```diff
@@ -0,0 +1,41 @@
...
+      path: /etc/default/grub
+      regexp: ^GRUB_CMDLINE_LINUX_DEFAULT=
+      line: GRUB_CMDLINE_LINUX_DEFAULT="quiet iommu=pt{% if '"GenuineIntel" in ansible_processor | unique' %} intel_iommu=on{% else %} amd_iommu=on{% endif %}{% if pve_mediated_devices_enabled %} i915.enable_gvt=1{% endif %}"
+    register: iommu_modify
+  - name: Update default grub
+    ansible.builtin.command: update-grub
+    when: iommu_modify.changed
+
+- name: Enable IOMMU unsafe interrupts
+  ansible.builtin.lineinfile:
```
lae, line 14: Use `ansible.builtin.copy` for this https://github.com/lae/ansible-role-proxmox/blob/v1.7.1/tasks/ipmi_watchdog.yml#L11 and specify the mode (0640 is fine)

tasks/pcie-passthrough.yml
```diff
@@ -0,0 +1,41 @@
...
+  - name: Modify modules list for GVT-g
+    ansible.builtin.blockinfile:
+      dest: /etc/modules
+      marker: "# {mark}: Modules required for GVT-g (managed by ansible)."
+      content: |
+        kvmgt
+    register: modules_list_gvtg
+    when: pve_mediated_devices_enabled
+  - name: Update initramfs
+    command: update-initramfs -u -k all
```
lae, line 40: Change to `ansible.builtin.command`

lae (changes requested):
This is a good start. Most of the requested changes are either for consistency or because `ansible-lint` gave a failure.

In addition to these changes, `kernel_module_cleanup.yml` should also be updated to add steps to cleanup changes when any of these role variables are unset or set to false. You can take a look at how it's done for other modules and mimick it for the most part.

lae:
Sorry for the late follow-up, I'm busy and haven't gotten a chance to run tests (takes forever) and just test it out in general, but this is on my radar and I'll let you know when I can.

There was one thing I was wondering whether or not to point out, but since it's probably gonna take me a while to actually get to this, I might as well. A lot of folks looking for VFIO in Proxmox are likely wanting it for their GPU(s), and in most cases that requires stubbing the GPU's vendor/product ID so that drivers don't try to drive them on the host. This PR is missing that and I was thinking it *could* be a separate follow-up PR, but since I think most folks need it anyway it could also be part of this one. It is explained in the PVE wiki a bit: https://pve.proxmox.com/wiki/Pci_passthrough#GPU_Passthrough

It may not be the best explanation (in fact, https://wiki.archlinux.org/title/PCI_passthrough_via_OVMF#Isolating_the_GPU is a better read), but as a real-world example this is my Proxmox workstation that I rely on to passthrough GPUs and USB controllers:

```
~$ cat /etc/modprobe.d/vfio.conf
softdep amdgpu pre: vfio-pci
softdep radeon pre: vfio-pci
options vfio-pci ids=1002:67df,1002:aaf0,1b73:1100,1002:6981,1002:aae0,1002:73e3 disable_vga=1
~$ grep vfio /etc/default/grub
GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX amd_iommu=on iommu=pt vfio-pci.ids=1002:67df,1002:aaf0,1b73:1100,1002:6981,1002:aae0,1002:73e3 pci.stub.ids=1002:67df,1002:aaf0,1002:6981,1002:aae0,1002:73e3"
```

Think adding a role variable+task for another `GRUB_CMDLINE_LINUX` line could work?

deloa6133:
Yes another line could handle this, I can take a look into it a bit more since I have not really had experience with dedicated GPU passhthrough, only integrated devices. I know this could require modifying/creating blacklists for drivers as well depending on device vendor/use case. Either way I can look into it a bit and possibly branch off of this PR to see if I can get something working properly

lae:
I don't think we have to worry too much about trying to cater to specific vendors in particular. Just having the user specify IDs and what drivers to use softdep on (I guess this is an alternative to actually blacklisting them) would get us most of the way, and if the user needs to do something else they can probably just add it to their playbook.

deloa6133:
Enabled pci id and driver vars that can configure default grub and vfio.conf file for gpu and other device passthrough. I think at this point a separate section in the README for `PCIe Passthrough:` would be appropriate to help people understand more in depth how this works. I'll write something up for that

lae:
I'll be available to review this properly mid-Oct. Please remind me then if I haven't.

ol3d:
Reminder to take a look into this PR more in depth

ol3d:
Closing due to inactivity, may re-open in the future

ol3d:
Apologies for the large amount of commits. Testing was done in my homelab and pulled from the repo instead of hosting the role locally. Over an extended peroid of time many changes/improvements were made. Everything should be current as of now. There are future enhancements I would like to add, however, I wanted to get this out for now for review.

PR #204, opened 2022-08-18
Idempotency fixes for proxmox_storage

### LINKED ISSUE #203: Present but disabled storage breaks idempotency
...or then I'm just setting some parameter the wrong way?

```
$ cat cluster.yml
pve_storages:
  - name: local
    type: dir
    path: /var/lib/vz
    disable: true
    state: present
```

Expected results:

```
RUN1
TASK [ansible-role-proxmox : Configure Proxmox Storage] *********************************************************************************************************************
changed: [server1] => (item={'name': 'local', 'type': 'dir', 'path': '/var/lib/vz', 'disable': True, 'state': 'present'})
RUN2
TASK [ansible-role-proxmox : Configure Proxmox Storage] *********************************************************************************************************************
ok: [server1] => (item={'name': 'local', 'type': 'dir', 'path': '/var/lib/vz', 'disable': True, 'state': 'present'})
```

Actual results:

```
RUN1
TASK [ansible-role-proxmox : Configure Proxmox Storage] *********************************************************************************************************************
changed: [server1] => (item={'name': 'local', 'type': 'dir', 'path': '/var/lib/vz', 'disable': True, 'state': 'present'})
RUN2
TASK [ansible-role-proxmox : Configure Proxmox Storage] *********************************************************************************************************************
changed: [server1] => (item={'name': 'local', 'type': 'dir', 'path': '/var/lib/vz', 'disable': True, 'state': 'present'})
```

lae:
Fixes #203 

Appears to at least fix the demonstrated issue under Vagrant.


library/proxmox_storage.py
```diff
@@ -202,9 +202,9 @@ def lookup(self):
         for item in self.existing_storages:
             if item['storage'] == self.name:
                 # pvesh doesn't return the disable param value if it's false,
-                # so we set it to False.
+                # so we set it to 0, which is what PVE would normally use.
                 if item.get('disable') is None:
-                    item['disable'] = False
+                    item['disable'] = 0
```
junousi, line 207: Unfortunately this still doesn't seem to resolve case from #203 

```
$ sha256sum library/proxmox_storage.py 
2e4962bce5833488f8fa8fcc89a87d4bb83b8368f99401ff9e25cd41e1c08dcf  library/proxmox_storage.py
$ ansible-playbook ...
...
TASK [ansible-role-proxmox : Configure Proxmox Storage] ***
changed: [server1] => (item={'name': 'local', 'type': 'dir', 'path': '/var/lib/vz', 'disable': True, 'state': 'present'})
```

Tried both variants `disable:yes` and `disable:true`. I'm starting to wonder if `/var/lib/vz` itself is a special case.

lae: I see you included a sha256sum and it matches, so never mind. Hm... I can't imagine `/var/lib/vz` being a special case though tbh. I'll try testing that out.

lae:
@junousi Are you specifying `content` for `/var/lib/vz`?

```
changed: [pve-3] => (item={'name': 'disabled-vz', 'type': 'dir', 'disable': True, 'path': '/var/lib/vz'}) => {
    "ansible_loop_var": "item",
    "changed": true,
    "item": {
        "disable": true,
        "name": "disabled-vz",
        "path": "/var/lib/vz",
        "type": "dir"
    },
    "state": "present",
    "updated_fields": [
        "content"
    ]
}
```

Going off of your play output I just used:

```
pve_storages;
  - name: disabled-vz
    type: dir
    disable: yes
    path: /var/lib/vz
```

And it's consistently showing up as changed as a result. In the module declaration `content` is supposed to be required and it is [defaulted to an empty list](https://github.com/lae/ansible-role-proxmox/blob/v1.7.1/tasks/main.yml#L275), but after some more testing it doesn't look like it's special treatment on `/var/lib/vz`; it just looks like PVE does this dumb thing where it returns a string `"none"` instead of a list when the content field is empty, so during comparison we're comparing an empty list to a string:

```
$ sudo pvesh get /storage/disabled-vz --output=json
{"content":"none","digest":"cd148351dd7079b10278343b930834dc318e959e","disable":1,"path":"/var/lib/vz","storage":"disabled-vz","type":"dir"}
```

Guess I'll add a clause for that in the lookup method.

lae:
Oh right, it's a comma-delimited string in the first place, not a list. I guess it's not as dumb as I thought.

junousi:
> @junousi Are you specifying content for /var/lib/vz?

Indeed I am not! No problem adding it wherever needed going forward.

Can do more testing next week earliest, though the updated logic does look sound.

junousi:
Can report that the latest changeset seems to fix the issue. Did a test for both enabled and disabled storage:

```
$ sha256sum roles/ansible-role-proxmox/library/proxmox_storage.py
d991905aa32c799d7a57fe3e57c77db2f14b7df181389a5a26ca6f788b2a1e0b  roles/ansible-role-proxmox/library/proxmox_storage.py

$ cat inventory/group_vars/cluster.yml
pve_storages:
  - name: local
    type: dir
    path: /var/lib/vz
    disable: false ##### enabled temporarily for testing
    state: present
$ ansible-playbook ... --tags storage # RUN1
TASK [ansible-role-proxmox : Configure Proxmox Storage] *********************************************************************************************************************
changed: [server1] => (item={'name': 'local', 'type': 'dir', 'path': '/var/lib/vz', 'disable': True, 'state': 'present'})
$ ansible-playbook ... --tags storage # RUN2
TASK [ansible-role-proxmox : Configure Proxmox Storage] *********************************************************************************************************************
ok: [server1] => (item={'name': 'local', 'type': 'dir', 'path': '/var/lib/vz', 'disable': True, 'state': 'present'})

$ cat inventory/group_vars/cluster.yml
pve_storages:
  - name: local
    type: dir
    path: /var/lib/vz
    disable: true ##### changed back
    state: present
$ ansible-playbook ... --tags storage # RUN1
TASK [ansible-role-proxmox : Configure Proxmox Storage] *********************************************************************************************************************
changed: [server1] => (item={'name': 'local', 'type': 'dir', 'path': '/var/lib/vz', 'disable': True, 'state': 'present'})
$ ansible-playbook ... --tags storage # RUN2
TASK [ansible-role-proxmox : Configure Proxmox Storage] *********************************************************************************************************************
ok: [server1] => (item={'name': 'local', 'type': 'dir', 'path': '/var/lib/vz', 'disable': True, 'state': 'present'})
```

PR #208, opened 2022-11-07
FIX new_kernel_exists in collect_kernel_info always returns true

pielas:
collect_kernel_info module reported kernel change(`new_kernel_exists` set to true) even in case when kernel did not change.

### DIFF
--- a/library/collect_kernel_info.py
+++ b/library/collect_kernel_info.py
@@ -34,7 +34,7 @@ def main():
             latest_kernel = kernel
 
     booted_kernel = "/lib/modules/{}".format(to_text(
-            subprocess.run(["uname", "-r"], capture_output=True).stdout.strip))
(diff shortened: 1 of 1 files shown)


lae:
Duplicate of #206? I was waiting for someone to review that and then forgot about it, actually...

pielas:
Ah, it looks so. I only checked if it's fixed on develop :)

lae:
Does that other PR look fine codewise and does it work for you, then? I think the `decode("utf-8")` isn't necessary because Ansible's `to_text()` method should take care of it, iirc.

pielas:
@lae yes, I wanted explicit conversion to string, but indeed it seems that it isn't necessary and for my case adding parentheses is enough

lae:
Ok thank you, I merged the other PR into develop, which I've actually been using for the past two months without issue (merged locally) now that I think about it.

PR #215, opened 2023-01-24
updated default ceph version to quincy

wu3rstle:
Because ceph-pacific will be EOL on 2023-06-01 the ceph version should be changed.


lae:
How does this affect existing ceph deployments? I think this will initiate an upgrade by default on existing clusters? Is that safe?

lae (approved):

PR #219, opened 2023-02-18
Fix host used to configure HA groups

btravouillon:
When the role is run with `--limit nodes_without_first_node`, the HA groups are not configured (tasks are skipped). Use `_init_node` instead of `groups[pve_group][0]` to run the tasks.

`_init_node` is any active node in an already initialized Proxmox cluster which is part of the run.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -208,7 +208,7 @@
     state: "{{ item.state | default('present') }}"
     comment: "{{ item.comment | default(omit) }}"
   with_items: "{{ pve_pools }}"
-  when: "not pve_cluster_enabled or (pve_cluster_enabled | bool and inventory_hostname == groups[pve_group][0])"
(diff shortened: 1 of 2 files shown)


lae:
Correct me if I'm wrong, but this looks like a possible QoL improvement throughout the entire role, actually, and not just for these tasks.

The tasks in `main.yml` that are run against the first cluster node probably could make use of this change. Do you think you can expand the scope of this to all tasks that reference `groups[pve_group][0]`?

lae:
@btravouillon since it's been some time, am I okay to assume your +1 is to indicate that you'll update this PR with what I suggested sometime?

btravouillon:
Oh yes, sorry. I'm quite busy this week but I'll try to take a look at this asap.

lae:
no problem, just checking. take your time.

lae (approved):

lae:
Looks like my vagrant setup is broken, but these changes seem trivial so I'll merge without testing.

PR #224, opened 2023-05-28
fix minimum host count in inventory group before enabling pve-cluster

tbasset:


lae (approved):
The default for `pve_cluster_enabled` is false so we should really be notifying the user that they're doing something wrong. We don't want to silently be overriding a user's configuration if it's bad.

I thought I had commented that before, but I guess not, and since it's been a while I'll just make that change.

PR #232, opened 2023-12-28
Add files for PVE 8 / Debian Bookworm

StrangelyTyped:
@zenntrix I have tested basic deployment on debian bookworm with this configuration, though not any of the additional features such as zfs or ceph.


zenntrix:
Thanks, I will test pve deployed Ceph and ZFS functionality.

PR #233, opened 2024-01-22
Patch upstream role on local fork

daanbosch:


nerdrobot (approved):
LGTM!

lae:
え？

FYI You probably want to use the `pve_manage_ssh` role variable instead.

PR #246, opened 2024-06-29
fix: set noninteractive for apt upgarde

jeschero:
Hey,

I have a little fix here that fixes the `apt upgrade` not waiting for user input anymore.
Otherwise it can happen that during `apt upgarde` apt will ask user something and since these are not answered in ansible, an error is generated.

Regards
jeschero

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -147,6 +147,8 @@
   retries: 2
   register: _dist_upgrade
   until: _dist_upgrade is succeeded
+  environment: 
+    - "DEBIAN_FRONTEND: noninteractive"
 
 - name: Perform system upgrades
   apt:


lae:
Should this be applied to all apt tasks or just the `dist-upgrade` one?

edv-pi:
to all apt tasks, theres no reason to use dialog for apt when run through ansible.

edv-pi (approved):

lae:
I am questioning whether this change does what it is intended to do. The `apt` module already seems to have this set:

https://github.com/ansible/ansible/blob/v2.17.1/lib/ansible/modules/apt.py#L1248

@jeschero Can you provide debug output for whatever playbook run was failing for you prior to making this change?

tasks/main.yml
```diff
@@ -147,6 +147,8 @@
   retries: 2
   register: _dist_upgrade
   until: _dist_upgrade is succeeded
+  environment: 
+    - "DEBIAN_FRONTEND: noninteractive"
```
lae, line 151: Syntax is off, anyway.

https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_environment.html#setting-the-remote-environment-in-a-task

lae (changes requested):

edv-pi:
ah lol, nice to know, didnt notice they added that

lae:
You're right that there's no reason for not using dialog when running apt through ansible, so I assumed that the module itself would already have disabled it in the first place instead of expecting users of the module to provide that env variable. And [it has been in ansible for 12 years](https://github.com/ansible/ansible-modules-core/commit/72aefdb609a16fa55903939b210d215f48925d75), apparently.

PR #247, opened 2024-07-02
VFIO/PCIe Passthrough configuration support

ol3d:
This is a new PR for vfio support within the lae.proxmox role. Closed out previous [PR](https://github.com/lae/ansible-role-proxmox/pull/202) due to sync issues between branches. This was to prevent force push commits in history and to keep history linear.


tasks/pcie_passthrough.yml
```diff
@@ -0,0 +1,110 @@
+---
+- name: Modify default grub configuration
+  notify: update-grub
+  block:
```
lae, line 4: Do you think we could use `blockinfile` for this? It'll just be a bunch of conditionals in the content block but should simplify removal into one task I think?

lae (changes requested):
Have not tested this, but this all looks pretty good.

tasks/pcie_passthrough.yml
lae: Actually, yeah, I do want this to be a block so that it'll be simpler for dropping in `amd_iommu` later.

lae:
fyi for future reference, since you mention keeping history linear, historically I prefer to have commits in a PR split into more logical chunks and without noise (lint, merges, commits with a nontrivial amount of code either removed or modified before merge) otherwise I'd squash the commit on merge, but since this has been a long running PR and would be quite a bit of work to rebase and refactor the commit history I'll keep it as-is.

lae (approved):

PR #251, opened 2024-07-07
Support for CIFS storages, namespacing in PBS, and ranking interfaces for corosync

edv-pi:
I added CIFS support and the new feature from pbs for namespaces so that it is possible to mount pbs namespaces with pve_cluster_prio0 and pve_cluster_prio1 it is possible to define which link has the higher priority.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -44,6 +44,8 @@ pve_cluster_clustername: "{{ pve_group }}"
 pve_manage_hosts_enabled: yes
 pve_cluster_addr0: "{{ ansible_default_ipv4.address if ansible_default_ipv4.address is defined else ansible_default_ipv6.address if ansible_default_ipv6.address is defined }}"
 # pve_cluster_addr1: "{{ ansible_eth1.ipv4.address }}
+# pve_cluster_addr0_priority: 0
+# pve_cluster_addr1_priority: 1
 pve_datacenter_cfg: {}
 pve_domains_cfg: []
 pve_cluster_ha_groups: []
@@ -57,4 +59,4 @@ pve_storages: []
 pve_ssh_port: 22
 pve_manage_ssh: true
 pve_hooks: {}
-pve_no_logging: false
+pve_no_log: false
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -57,4 +57,4 @@ pve_storages: []
 pve_ssh_port: 22
 pve_manage_ssh: true
 pve_hooks: {}
-no_logging: false
+pve_no_logging: false
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -57,3 +57,4 @@ pve_storages: []
 pve_ssh_port: 22
 pve_manage_ssh: true
 pve_hooks: {}
+no_logging: false
--- a/library/proxmox_storage.py
+++ b/library/proxmox_storage.py
@@ -103,7 +103,22 @@ options:
         description:
             - Specifies whether or not the given path is an externally managed
             mountpoint.
-
+    namespace:
+        required: false
+        description:
+            - Specifies the Namespace that should be used on PBS
+    share:
+        required: false
+        description:
+            - Specifies the CIFS-Share to use
+    subdir:
+        required: false
+            - specifies the folder in the share dir to use for proxmox 
+              (useful to seperate proxmox content from other content)
+    domain:
+        required: false
+            - Specifies Realm to use for NTLM/LDAPS Authentification if using 
+              an AD-Enabled share
 author:
     - Fabien Brachere (@fbrachere)
 '''
@@ -170,6 +185,7 @@ EXAMPLES = '''
     datastore: main
     fingerprint: f2:fb:85:76:d2:2a:c4:96:5c:6e:d8:71:37:36:06:17:09:55:f7:04:e3:74:bb:aa:9e:26:85:92:63:c8:b9:23
     encryption_key: autogen
+    namespace: Top/something
 - name: Create a ZFS storage type
   proxmox_storage:
     name: zfs1
@@ -177,6 +193,17 @@ EXAMPLES = '''
     content: [ "images", "rootdir" ]
     pool: rpool/data
     sparse: true
+- name: CIFS-Share
+  proxmox_Storage:
+    name: cifs1
+    server: cifs-host.domain.tld
+    type: cifs
+    content: ["snippets", "vztmpl", "iso" ]
+    share: sharename
+    subdir: /subdir
+    username: user
+    password: supersecurepass
+    domain: addomain.tld
 '''
 
 RETURN = '''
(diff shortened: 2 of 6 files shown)


defaults/main.yml
```diff
@@ -57,3 +57,4 @@ pve_storages: []
 pve_ssh_port: 22
 pve_manage_ssh: true
 pve_hooks: {}
+no_logging: false
```
lae, line 60: Role variable should be prefixed with `pve_`. Given how this variable is used, I'd probably just use `pve_no_log` instead of `pve_no_logging`.

library/proxmox_storage.py
```diff
@@ -170,13 +185,25 @@
...
     type: zfspool
     content: [ "images", "rootdir" ]
     pool: rpool/data
     sparse: true
+- name: CIFS-Share
+  proxmox_Storage:
+    name: cifs1
+    server: cifs-host.domain.tld
+    type: cifs
+    content: ["snippets", "vztmpl", "iso" ]
```
lae, line 201: needs a space before `"snippets"`

library/proxmox_storage.py
```diff
@@ -170,13 +185,25 @@
...
+    namespace: Top/something
 - name: Create a ZFS storage type
   proxmox_storage:
     name: zfs1
     type: zfspool
     content: [ "images", "rootdir" ]
     pool: rpool/data
     sparse: true
+- name: CIFS-Share
+  proxmox_Storage:
```
lae, line 197: storage should not be capitalized

library/proxmox_storage.py
```diff
@@ -220,7 +247,10 @@ def __init__(self, module):
         self.thinpool = module.params['thinpool']
         self.sparse = module.params['sparse']
         self.is_mountpoint = module.params['is_mountpoint']
-
```
lae, line 223: newline should be kept (and after your inserts/before the `# Validate ...` comment), it's here for readability.

library/proxmox_storage.py
```diff
@@ -305,16 +335,26 @@ def prepare_storage_args(self):
...
+        if self.subdir is not None:
+            args['subdir'] = self.subdir
+        if self.domain is not None:
+            args['domain'] = self.domain
+        # end cifs
         if self.maxfiles is not None and 'backup' not in self.content:
             self.module.fail_json(msg="maxfiles is not allowed when there is no 'backup' in content")
         if self.krbd is not None and self.type != 'rbd':
             self.module.fail_json(msg="krbd is only allowed with 'rbd' storage type")
-
```
lae, line 317: Keep newline before `return` statement.

library/proxmox_storage.py
```diff
@@ -386,7 +426,7 @@ def main():
         nodes=dict(type='list', required=False, default=None),
         type=dict(default=None, type='str', required=True,
                   choices=["dir", "nfs", "rbd", "lvm", "lvmthin", "cephfs",
-                           "zfspool", "btrfs", "pbs"]),
+                           "zfspool", "btrfs", "pbs","cifs"]),
```
lae, line 429: Needs space after comma.

library/proxmox_storage.py
```diff
@@ -405,7 +445,12 @@ def main():
         vgname=dict(default=None, type='str', required=False),
         thinpool=dict(default=None, type='str', required=False),
         sparse=dict(default=None, type='bool', required=False),
-        is_mountpoint=dict(default=None, type='bool', required=False),
+        is_mountpoint=dict(default=None, type='bool', required=False),        
+        namespace=dict(default=None, type='str', required=False),
+        subdir=dict(default=None, type='str', required=False),
+        domain=dict(default=None, type='str', required=False),
+        share=dict(default=None, type='str', required=False),
+
```
lae, line 453: Remove newline.

tasks/main.yml
```diff
@@ -333,6 +333,12 @@
     vgname: "{{ item.vgname | default(omit) }}"
     thinpool: "{{ item.thinpool | default(omit) }}"
     sparse: "{{ item.sparse | default(omit) }}"
+    namespace: "{{ item.namespace | default(omit) }}"
+    domain: "{{ item.domain | default(omit) }}"
+    subdir: "{{ item.subdir | default(omit) }}"
+    share: "{{ item.share | default(omit) }}"
+# Set logging to true in production in ypur groupvars
```
lae, line 340: This comment can be left out, or rather moved into the documentation.

tasks/main.yml
```diff
@@ -333,6 +333,12 @@
     vgname: "{{ item.vgname | default(omit) }}"
     thinpool: "{{ item.thinpool | default(omit) }}"
     sparse: "{{ item.sparse | default(omit) }}"
+    namespace: "{{ item.namespace | default(omit) }}"
+    domain: "{{ item.domain | default(omit) }}"
+    subdir: "{{ item.subdir | default(omit) }}"
+    share: "{{ item.share | default(omit) }}"
+# Set logging to true in production in ypur groupvars  
+  no_log: "{{ no_logging | default(false) }}"
```
lae, line 341: `default(false)` is redundant here as it's already specified in a defaults file. Should remove it and update the role variable to the new name.

tasks/pve_add_node.yml
```diff
@@ -17,9 +17,9 @@
 - name: Add node to Proxmox cluster
   ansible.builtin.command: >-
     pvecm add {{ hostvars[_init_node].pve_cluster_addr0 }} -use_ssh
-    -link0 {{ pve_cluster_addr0 }}
+    -link0 {{ pve_cluster_addr0 }},priority={{ pve_cluster_prio0 | default(omit)}}
```
lae, line 20: `default(omit)` I believe breaks here? It's meant for omitting certain attributes of a module and is meant to be the only value in the attribute field, e.g.

```yaml
ansible.builtin.module:
  attribute: "{{ some_var | default(omit) }}"
```

IIRC technically this sets `attribute` to something like `omit-randomhash` and Ansible interprets this to ignore the attribute. The way it's being used here, I believe it'll just insert something like `-link0 1.2.3.4,priority=omit-e52bf6d0bf43` and use that as-is.

Either way, it looks like what you want to do is omit the `priority` value anyway but that's hardcoded into the string so it wouldn't be left out in the first place. This should be something like `-link0 {{ pve_cluster_addr0 }}{{ ",priority=" + pve_cluster_prio0 if pve_cluster_prio0 is defined }}`, but I'm not sure if this is valid syntax in Ansible's Jinja2 templates. `{% if pve_cluster_prio0 is defined %},priority={{ pve_cluster_prio0 }}{% endif %}` is a bit longer but should work I think, if the other one doesn't.

lae (changes requested):
Thanks for the history cleanup.

In addition to the comments, can you add documentation for the new role variables as well as the CIFS storage support (this should go under `Storage Management`) to the README?

Also, I'm not sure what value is supported to go into `pve_cluster_prio0` etc (documentation should help here), but if I'm assuming correctly that it's an integer for ranking that particular interface, then I *think* `pve_cluster_prio0` etc should really be `pve_cluster_addr0_priority` (and same for `prio1`), otherwise it gives me the impression that `prio0` is yet another interface and not associated to `addr0`.

library/proxmox_storage.py
edv-pi: added the newline but github seems not to detect the change :/

lae (approved):
Weird that the PVE docs doesn't mention namespaces (it's a valid flag, just seems undocumented).
https://github.com/proxmox/pve-docs/blob/f5e17ae1c680a475e980230d4d96a36db2c7afcd/pve-storage-pbs.adoc

Anyway, I added some additional changes since there were inconsistencies. Also for future reference, links don't work in code blocks (so the one for priority was invisible in the generated document).

PR #255, opened 2024-08-23
Fix all ansible-lint and yamllint issues

thystips:


thystips:
Juste in case : I have not tested playbooks in `test/` directory.

lae (changes requested):
I'll need more time to finish this review later but for the time being there are some things I need feedback on. Please also rebase against develop (and amend the commit).

tasks/ceph.yml
```diff
@@ -58,80 +60,87 @@
...
+            path: crush_map_decompressed
+            regexp: >-
+              rule\s+{{ item.name }}\s+{(?:(?P<space>\s+)id\s+(?P<id>[^\s]+)|\s+type\s+(?P<r_type>[^\s]+)|\s+min_size[ ](?P<min>[^\s]+)|\s+max_size\s+(?P<max>[^\s]+)|\s+step\s+take\s+default(?:\n|\s+class\s+(?P<class>[^\n]*))|\s+step\s+(?P<choose>chooseleaf|choose).*?type\s+(?P<type>[^\s]+))+(?:.|\n)*?}
+            replace: "rule {{ item.name }} {\
+              \\g<space>id \\g<id>\
+              \\g<space>type \\g<r_type>\
+              \\g<space>min_size {{ (pve_ceph_crush_rules | selectattr(\"name\", \"match\", item.name) | list)[0].min_size | default(\"\\g<min>\") | trim }}\
+              \\g<space>max_size {{ (pve_ceph_crush_rules | selectattr(\"name\", \"match\", item.name) | list)[0].max_size | default(\"\\g<max>\") | trim }}\
+              {%- if ((pve_ceph_crush_rules | selectattr(\"name\", \"match\", item.name) | list)[0].class | default(False)) -%}\
               \\g<space>step take default class {{ (pve_ceph_crush_rules | selectattr(\"name\", \"match\", item.name) | list)[0].class }}\
```
lae, line 103: this and the untouched line below should also be indented for readability (because they're inside conditionals)

tasks/ceph.yml
```diff
@@ -58,80 +60,87 @@
...
+              \\g<space>step emit\n}"
+          loop: "{{ pve_ceph_crush_rules }}"
+          register: _crushmap
+          # yamllint enable rule:line-length
+
+        - name: Validate and compress new crushmap
+          ansible.builtin.command: crushtool -c crush_map_decompressed -o new_crush_map_compressed
+          register: _crushmap_valid
+          when: _crushmap.changed
+          changed_when: _crushmap_valid.rc == 0
```
lae, line 117: This should probably just be true? Cause the other cases are just task error if rc != 1 or task skip...I think? If that's accurate then this statement just feels like it's an unnecessary layer of indirection cause it'll always be successful as long as it completes.

tasks/ceph.yml
```diff
@@ -155,97 +164,101 @@
...
         pveceph fs create
         --name {{ item.name }}
         --add-storage {{ item.storage }}
         --pg_num {{ item.pgs }}
       register: _ceph_fs_create
       failed_when: _ceph_fs_create.stderr
       when: "item.name not in (_ceph_fs.stdout | from_json | map(attribute='name'))"
-      with_items: '{{ pve_ceph_fs }}'
+      with_items: "{{ pve_ceph_fs }}"
+      changed_when: _ceph_fs_create.rc == 0
```
lae, line 206: Same thing about the layer of indirection here.

tests/test.yml
lae, line 1: Can you add a comment to this file stating that all the shell tasks were replaced with command tasks?
As you've said you haven't tested these so there may or may not be breakage here...I don't really think there is, but just in case, that comment will help me remember if I run into any snags when I get around to rebuilding the test environment.

lae (changes requested):
Okay, done checking the last two files. See comments and last review.

PR #258, opened 2024-08-26
feat: Add IOMMU for AMD CPU

thystips:
Add AMD IOMMU support in Grub Arguments

### DIFF
--- a/tasks/pcie_passthrough.yml
+++ b/tasks/pcie_passthrough.yml
@@ -5,6 +5,7 @@
     marker: "# {mark}: IOMMU default grub configuration (managed by ansible)."
     content: "\
       {% if '\"GenuineIntel\" in ansible_processor | unique' %}GRUB_CMDLINE_LINUX=\"$GRUB_CMDLINE_LINUX intel_iommu=on\"\n{% endif %}\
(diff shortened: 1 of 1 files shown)


lae:
This was supposedly not necessary according to https://github.com/lae/ansible-role-proxmox/pull/247 (https://github.com/lae/ansible-role-proxmox/pull/247/commits/f4dcad246dc1a7519204950a8acc9b4f0bbdda3c), though I hadn't yet gotten a chance to validate it. Is it inaccurate?

thystips:
My bad, after another amd_iommu isn't needed with AMD because is enable by default. Is only require for Intel.
This is an error present on several of my servers.

lae:
sorry @thystips, I didn't quite understand your message. Yes, the commit I linked states that it's enabled by default, but I haven't checked if this is true. Do you mean to say that you confirmed that it is? And if so, then what error are you talking about?

thystips:
After your message I've checked on Proxmox forum and IOMMU for AMD is automatically enabled: https://forum.proxmox.com/threads/is-iommu-enabled-on-my-system.128191/post-560919, so in my case I only `iommu=pt` but this already in this role.

But I work with an old playbook that I didn't write and there is a specific task (redundant with this role) to edit kernel parameters based on CPU vendor and we add amd_iommu but is not necessary. There is no error on your Ansible role.

PR #262, opened 2024-08-30
Metric Servers management in Proxmox

### LINKED ISSUE #259: Feature Request: Add Support for Managing Proxmox Metric Servers
I would like to request the addition of support for managing Proxmox Metric Servers within this role.

One possible approach to implement this feature would be to create a template for the /etc/pve/status.cfg file and restart the pvestatd service. However, this solution does not allow for the management of passwords credentials.

An alternative and more secure method would be to add support using the Proxmox API, which would allow for full management, including password handling. Unfortunately, I am less familiar with this approach, so I'm unable to propose the code myself.

Thank you for considering this feature!

thystips:
Tested only with PVE 7

Telegraf can be used to transfert metrics to Prometheus, Elasticsearch, etc.

Resolve #259

### DIFF
--- a/library/proxmox_metric_server.py
+++ b/library/proxmox_metric_server.py
@@ -1,0 +1,411 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+ANSIBLE_METADATA = {
+    "metadata_version": "0.1",
+    "status": ["preview"],
+    "supported_by": "thystips",
+}
+
+DOCUMENTATION = """
+---
+module: proxmox_metric_server
+short_description: Manages the Metric Server in Proxmox
+options:
+    id:
+        required: true
+        type: str
+        description:
+            - ID of the server.
+    port:
+        type: int
+        default: 8089
+        description:
+            - Port of the server.
+    server:
+        required: true
+        type: str
+        description:
+            - Server dns name or IP address.
+    type:
+        type: str
+        default: "influxdb"
+        choices: [ "influxdb", "graphite" ]
+        description:
+            - Plugin type.
+    protocol:
+        type: str
+        default: "udp"
+        choices: [ "udp", "tcp", "http", "https" ]
+        description:
+            - Protocol used to send metrics.
+            - "http" and "https" are only available for "influxdb" type and using http v2 api.
+            - "tcp" is only available for "graphite" type.
+    disable:
+        type: bool
+        default: false
+        description:
+            - Disable the metric server.
+    organization:
+        type: str
+        description:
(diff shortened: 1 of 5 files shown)


library/proxmox_metric_server.py
```diff
@@ -0,0 +1,411 @@
...
+        self.path = module.params["path"]
+        self.api_path_prefix = module.params["api_path_prefix"]
+        self.timeout = module.params["timeout"]
+        self.max_body_size = module.params["max_body_size"]
+        self.mtu = module.params["mtu"]
+        self.verify_certificate = module.params["verify_certificate"]
+        self.state = module.params["state"]
+
+        try:
+            self.existing_server = pvesh.get("cluster/metrics/server")
```
lae, line 149: minor nitpick, but pluralize var?

library/proxmox_metric_server.py
```diff
@@ -0,0 +1,411 @@
...
+
+        self.parse_servers()
+
+    def parse_servers(self):
+        self.servers = []
+        for existing_server in self.existing_server: # type: ignore
+            self.servers.append(existing_server.get("id"))
+
+    def lookup(self):
+        self.servers = []
```
lae, line 161: unused variable

lae (changes requested):
Overall very well done. Thanks for the PR!

There's one critical change needed (the imports) and I've left comments for some other things.

lae:
oh, feel free to add your name to the list of contributors for metric server configuration support. (otherwise I'll add it later)

thystips:
I initially write this module for a private collection where this module is used as standalone after proxmox install, I forgot edit some code like import. It' s fixed.
All requests solved.

library/proxmox_metric_server.py
```diff
@@ -340,10 +338,9 @@ def main():
             msg="The 'protocol' parameter must be 'tcp' or 'udp' for 'graphite' type."
         )
 
-    if module.params["type"] == "graphite" and (
-        module.params["bucket"] is not None
-        or module.params["organization"] is not None,
-        module.params["token"] is not None,
+    if module.params["type"] == "graphite" and any(
```
lae, line 341: I'm assuming it's intentional, but the conditions in this statement slightly changed ( `a || b && c` → `a || b || c` ).

lae (approved):

library/proxmox_metric_server.py
thystips: Yes it is intentional. Graphite have less arguments in API.

PR #276, opened 2024-11-09
Add the abillity to set extra pubkeys in root`s authorized_keys via this role

virtualized-human:
Hello,

with these changes (which I may have to change again if it is not perfect, but it works as expected with 3 node cluster as well as non-cluster node with own infra tested) I have implemented the possibility to manage the authorized_keys via optional variable via this role.

As long as the var `pve_ssh_extra_keys` is not set, not much will change.

Where I have changed something else is that I have moved the whole authorized_keys part to a new tasks file. The reason is that I don't quite understand why this is limited to clusters and single node constructs can't work with the variable `pve_ssh_extra_keys` and any other ssh management stuff in the future. If you don't want that (ssh management) you can still set `pve_manage_ssh` to false.

Now to what I have tested. So I just used the `exclusive` specification, which is only set to true if `pve_ssh_extra_keys` is set and longer than 0 (i.e. contains something) to make sure that it does not accidentally interfere with existing setups (assuming you manage authorized_keys with another role). In the defaults, the variable is set empty (= length 0).

Here is a visual comparison:

extra_keys_disabled
extra_keys_enabled

Criticism/improvements very welcome! I really want this feature and use it myself now. Because the implementation of an additional role is a bit more difficult in this case (if you want to make sure that there are really only exactly the proxmox own keys + defined ones and manual added keys get overwritten).

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,6 +227,11 @@ This is useful if you want to use InfluxDB, Graphite or other (with telegraf).
 a port other than the default 22, please set this variable. If a new node is
 joining the cluster, the PVE cluster needs to communicate once via SSH.
 
+`pve_ssh_extra_keys` allows you to add additional SSH public keys to the root
+users SSH authorized_keys, which are declaratively managed, so that keys that
+are not specified anymore, will get removed... requires `pve_groups` to be set
+right and `pve_manage_ssh` to be true.
+
 `pve_manage_ssh` (default true) allows you to disable any changes this module
 would make to your SSH server config. This is useful if you use another role
 to manage your SSH server. Note that setting this to false is not officially
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -59,6 +59,7 @@ pve_acls: []
 pve_storages: []
 pve_metric_servers: []
 pve_ssh_port: 22
+pve_ssh_extra_keys: []
 pve_manage_ssh: true
 pve_hooks: {}
 pve_no_log: false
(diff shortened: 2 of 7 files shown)


lae:
> The reason is that I don't quite understand why this is limited to clusters and single node constructs can't work with the variable pve_ssh_extra_keys and any other ssh management stuff in the future. If you don't want that (ssh management) you can still set pve_manage_ssh to false.

This expands the scope of this role beyond what it should from my point of view. There are other Ansible roles for managing SSH keys that people should be using instead if they need SSH access to Proxmox nodes as root (and thus with `pve_manage_ssh` set to false). The only reason authorized keys are touched at all is to allow clustering to happen, nothing more.

virtualized-human:
okay. and how exactly should the external role differentiate between deleting and the keys that originate from this role? I see it as the task of this role to implement the pubkeys of the nodes, as is currently the case. But how does an external role distinguish between these and those that have been deliberately added manually? The role does not currently give me a single indicator by which I could recognize this externally?

my only conceivable external solution would be to completely dispense with your ssh implementation, make a copy paste of the ssh tasks and implement there what is included in my commits here. but with an external role, I am again taking over tasks that should actually take over your role or currently already does, but I still don't know how an external role is supposed to distinguish between your keys and others. however, if I implement all this in this role, ansible knows exactly what comes from where.

Do you have any ideas for me?

lae:
Setting `pve_manage_ssh` to false would prevent this role from generating any keys in the first place, so there would be no need to distinguish SSH keys that originate from this role because...they don't exist? (at least if you're applying against a new set of Debian machines and not toggling that afterwards) You would just have that external role create or place an SSH key for the root user which gets added to authorized_keys for all the nodes in a cluster. Unless I'm misunderstanding whatever issue you're running into?

There's not really a need to copy paste *all* of the SSH tasks, half of them are just to restrict exactly how and when the generated SSH keys are used, which is more of a nice to have (principle of least privilege) than a necessity. As long as `ssh root@x.y.z.node` works from every other node, without a password/host key verification prompt, then `pvecm` will proceed as normal.

PR #302, opened 2025-06-18
Support amd passthrough

blake-hamm:
I have a proxmox host with an AMD cpu. I discovered that the IOMMU grub were not updating properly. This change addresses that and also fixes logic which would always be true.


blake-hamm:
> If you can rename the PR and amend your commits (new PR is not necessary) to cleanup `intel_iommu` getting inserted on AMD systems I'll at least merge that bit in.

Updated! Thanks for the callout.. I'll search the issues next time :melting_face:

PR #311, opened 2025-08-04
feat(library) add proxmox_sdn

Derioss:
Thanks for your module, we added SDN features for our needs

### DIFF
--- a/library/proxmox_query.py
+++ b/library/proxmox_query.py
@@ -1,12 +1,12 @@
 #!/usr/bin/python
 
 ANSIBLE_METADATA = {
-    'metadata_version': '1.0',
-    'status': ['stableinterface'],
-    'supported_by': 'lae'
+    "metadata_version": "1.0",
+    "status": ["stableinterface"],
+    "supported_by": "lae",
 }
 
-DOCUMENTATION = '''
+DOCUMENTATION = """
 ---
 module: proxmox_query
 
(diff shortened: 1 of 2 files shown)


lae (changes requested):
The `proxmox_query` is really only meant to be a diagnostic module (i.e. a module for collecting facts from the cluster to support other tasks), and not one for configuring PVE. And given there appear to already be `set /cluster/sdn/?/?` calls in `proxmox_sdn.py`, I'm not sure why the `set /cluster/sdn` is being separated out of it in 6a4f28c9c38eb0c188f32f9e9c20fb9aa217b4b3.

There also doesn't appear to be a way to use the `proxmox_sdn` module via the role in the current changeset. Please add the relevant task for managing it via role variables and documentation in the README. Since this is a big feature it should probably have its own section. See https://github.com/lae/ansible-role-proxmox/commit/a42e6b5814d094f0547d125b43be1fe51f5ce6d6 for a decent (but simpler than this) example of adding a new management module.

Derioss:
I admit I've enhanced the query module because I use it for other purposes, but you're right that it's not logical to integrate the SDN application within the module itself. I should have a time slot available for testing in early September.

lae:
I'm not sure if you're the original author of these changes (the commits don't appear to belong to your account) and given the lack of movement, I'll close this for now. If you intend to return to this PR, reply to this thread before you make any changes to your branch so I can reopen the PR (don't open a new PR, please).

PR #327, opened 2025-10-17
Ceph Erasure Coding

commonism:
Ceph added [EC](https://docs.ceph.com/en/latest/rados/operations/erasure-code/) in [2017](https://ceph.io/en/news/blog/2017/new-luminous-erasure-coding-rbd-cephfs/).

These changes add the ability to configure a storage pool as EC pool.

> This has the potential to dramatically reduce the overall cost per terabyte of Ceph systems since the usual 3x storage overhead of replication can be reduced to more like 1.2x to 2x, depending on the choice of erasure code.

### DIFF
--- a/library/proxmox_storage.py
+++ b/library/proxmox_storage.py
@@ -70,7 +70,7 @@ options:
         description:
             - Ceph/ZFS pool name.
             - For erasure-coded (EC) pools, with a data and metadata pool, enter the metadata pool name here.
-    datapool:
+    data_pool:
         required: false
         type: str
         description:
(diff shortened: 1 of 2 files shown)


commonism:
The origin of the changes is https://github.com/cispa/ansible-role-proxmox/.
I rebased them to get them merged.

lae (changes requested):
Since the flag passed to the API is `data-pool`, can you rename the attribute to `data_pool` so it's consistent with the other flags using hyphens?

PR #340, opened 2025-11-19
Workflow to publish to Ansible Galaxy

lae:


gilou:
I guess the only way to try it.. is to get a new release in no? :P
The new ansible 2.20 deprecations could be the flag for it.. as v1.10.0 makes it rather unhappy hehe.. did you push v1.10.0 manually btw?

lae:
> did you push v1.10.0 manually btw?

yes

gilou:
I don't know much about github actions, so can't seriously review that, and to test it would be weird I guess, or I'd had to set it up on my own namespace / account.. so as unhelpful as my proposal is.. let's merge that, so that 1.10.1 or 1.11.0 (not sure what's your angle there, I'd do 1.11 for ansible 12 / 2.20 support..)

lae:
time to click the red bypass checkbox and merge i guess

PR #343, opened 2025-12-10
"Fix: Add missing '| bool' filter to pve_cluster_enabled in groups task

pofo14:
The groups task condition was missing the '| bool' filter on the first
pve_cluster_enabled check, causing it to evaluate incorrectly when the
variable is a string. This brings it in line with the roles and users
tasks which already have this filter.

The issue was the groups condition was always evaluating to false.  Therefore my user creation was failing as the group did not exist.  The condition was different from both the roles and user condition, and this brings it to be the same.

After this change the groups were created as expected.


lae (approved):
:+1:

PR #345, opened 2026-01-02
Auto install the package python3-debian needed by the module deb822

PendaGTP:
Currently, to run this role, one host requirement is `python`, which is fine

But there is an issue when using Ansible > 12 because the module `ansible.builtin.deb822_repository`  requires `python3-debian`

Using `pve_extra_packages` will not works since `deb822_repository` is used before so it would be nice to use the [install_python_debian](https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/deb822_repository_module.html#parameter-install_python_debian) parameter to automatically handle this dependency for this module

If you think this requirement should not be handled by this role or by this way, let me know and/or close this PR

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -164,6 +164,7 @@
         components: "{{ pve_repository.components }}"
         signed_by: /etc/apt/trusted.gpg.d/proxmox-release-{{ ansible_distribution_release }}.gpg
         state: present
+        install_python_debian: true
       register: _pve_repo
 
     - name: Add Proxmox Ceph repository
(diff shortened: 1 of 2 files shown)


lae:
seems like you closed this yourself and I don't see these changes in your other PR—did this become irrelevant?

PendaGTP:
I'm not sure if we need to ensure that the package python3-debian is installed by directly using the param install_python_debian. I haven't yet tested if this package is also required by another task, so that's why I have close this PR.

I don't know which would be better:

- no installing it, letting the user install requirements (python and python3-debian) before using the role
- or installing it at the beginning of the role as requirement

Let me know, and I can create a new PR (to add a note in the docs or add an earlier task for the python3-debian requirement), once decided on the desired behavior

PR #155, opened 2021-09-07
Feature/pve7

zenntrix:

PR #347, opened 2026-01-03
Fix repository management issues and simplify configuration

### LINKED ISSUE #342: Ceph repositories are removed with Proxmox 9
I noticed the Ceph repositories are created then removed in the same run.
```
TASK [lae.proxmox : Add Proxmox Ceph repository] ******************************************************************************************************************************************************************
Wednesday 19 November 2025  21:48:24 +0000 (0:00:00.321)       0:00:18.478 ****
Wednesday 19 November 2025  21:48:24 +0000 (0:00:00.321)       0:00:18.478 ****
changed: [bb01]
changed: [bb02]
changed: [bb03]

[...]
TASK [lae.proxmox : Remove automatically installed PVE Enterprise repo configuration] *****************************************************************************************************************************
Wednesday 19 November 2025  21:45:22 +0000 (0:00:00.078)       0:00:17.972 ****
Wednesday 19 November 2025  21:45:22 +0000 (0:00:00.078)       0:00:17.971 ****
ok: [bb01] => (item={'name': 'pve-enterprise', 'uris': 'https://enterprise.proxmox.com/debian/pve', 'components': ['pve-enterprise']})
ok: [bb02] => (item={'name': 'pve-enterprise', 'uris': 'https://enterprise.proxmox.com/debian/pve', 'components': ['pve-enterprise']})
ok: [bb03] => (item={'name': 'pve-enterprise', 'uris': 'https://enterprise.proxmox.com/debian/pve', 'components': ['pve-enterprise']})
changed: [bb01] => (item={'name': 'ceph', 'uris': 'https://enterprise.proxmox.com/debian/no-subscription', 'components': ['enterprise']})
changed: [bb02] => (item={'name': 'ceph', 'uris': 'https://enterprise.proxmox.com/debian/no-subscription', 'components': ['enterprise']})
changed: [bb03] => (item={'name': 'ceph', 'uris': 'https://enterprise.proxmox.com/debian/no-subscription', 'components': ['enterprise']})
```

I'm using the default `pve_ceph_repository` which creates this configuration:
```
$ cat /etc/apt/sources.list.d/ceph.sources
Components: no-subscription
X-Repolib-Name: ceph
Signed-By: /etc/apt/trusted.gpg.d/proxmox-release-trixie.gpg
Suites: trixie
Types: deb
URIs: http://download.proxmox.com/debian/ceph-squid
```

I guess the `state: absent` may only consider `name` to select which repository to remove.

https://github.com/lae/ansible-role-proxmox/blob/a83346eff8ee4e63f972dd87078ed9369e000658/tasks/remove_enterprise_repos.yml#L14-L25

PendaGTP:
Fixes #342

This PR fixes and improves how the role manages Proxmox repositories

### Main Fixes

- Fixed `remove_enterprise_repos.yml` always removing the Ceph repository when using deb822 format
- Fixed `pve_ceph_debian_component` being incorrectly used for the Ceph repository, which led to an invalid repository URL (see: https://github.com/lae/ansible-role-proxmox/issues/342#issuecomment-3555235083)

### Improvements

-  ~~Simplifies repository configuration by unifying config with existing `pve_repository` and `pve_ceph_repository` variables ( removes `pve_repository_line` and `pve_ceph_repository_line`)~~
- Ensures compatibility of Ceph repository between PVE 9 and bellow (handles filename migration from `.list` to `.sources`)

Note:

- Preserves backward compatibility with old PVE releases
- Preserves backward compatibility with old Debian releases
- There is no separate "Remove Ceph repository" task because, unlike the PVE repository, the Ceph repository uses the same filename (`ceph`) for both enterprise and non-enterprise versions, so the logic differs from PVE repository handling

### ~~Breaking Changes~~

~~**Users with default settings (no subscription):**~~

~~- No action required. The role will continue to work with default settings~~

~~**Users with test or enterprise subscriptions:**~~

~~- Must update `pve_repository` and/or `pve_ceph_repository` variables if they were previously using `pve_repository_line` and `pve_ceph_repository_line`~~

EDIT: preserve `pve_x_repository_line` format, no breaking changes


lae (changes requested):
per https://github.com/lae/ansible-role-proxmox/pull/314#issuecomment-3162161668 (and previous comments in that thread) `pve_repository_line` needs to remain for now, with maybe modification to the logic as to how it's used (i.e. dissect the components into `pve_repository` for use in the config task).

PendaGTP:
Thanks for your feedback!

 I understand that removing `pve_repository_line` is not planned for now, I will update the pr to maintain both repository configuration format

PendaGTP:
Done, let me know if any other changes are needed

lae:
busy with https://www.youtube.com/watch?v=IrbjHrMaOwg, will look later

will note that `pve_repository_line` is always defined with recent changes, dunno if intentional or hallucination

PendaGTP:
Oh I think I didn't understand correctly your previous comment, I thought that we need to keep `pve_repository_line` and `pve_ceph_repository_line`

So my changes, revert thems but I added a "translation" on the task to only use `pve_repository` and `pve_ceph_repository` format to add/remove repositories

What it does:

- if using x_repository_line but not x_repository them convert
- if using x_repository then use it

Maybe I completly miss something and/or maybe you mean that we need to remove the defaults `pve_repository_line` and `pve_ceph_repository_line` or comment them?

lae:
I think you understood the previous comment. But keeping it in the defaults means `pve_repository_line` is always defined, so it'll always replace `pve_repository` even if the user defines `pve_repository`. So removing the default would be the correct thing to do.

PendaGTP:
Aaaah yes mb, I was thinking the user would set pve_repository_line to empty but that's obviously not ideal

let me know if requires other changes

PendaGTP:
since this was a little bit related, fyi: https://github.com/lae/ansible-role-proxmox/pull/345#issuecomment-3707948150

tasks/repositories.yml
```diff
@@ -0,0 +1,103 @@
+---
+- name: Convert pve_repository_line to pve_repository format
+  ansible.builtin.set_fact:
+    pve_repository:
+      uris: "{{ (pve_repository_line | trim | split(' '))[1] }}"
+      suites: "{{ (pve_repository_line | trim | split(' '))[2] }}"
+      components: "{{ (pve_repository_line | trim | split(' '))[3:] | join(' ') }}"
```
lae: it looks like the `components` attribute is a list [according to the documentation](https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/deb822_repository_module.html), so a `join` shouldn't be necessary here.

and on a similar note, I am not sure if it accepts a string as a single item list or if it tries to parse that string as a list of characters. I'm assuming it's fine because no one has complained about the default so far, but I would like to confirm that first. If it is the latter the default might need to be updated.

tasks/repositories.yml
```diff
@@ -0,0 +1,103 @@
...
+  when:
+    - ansible_distribution_major_version | int >= 12
+
+- name: Add PVE repository (legacy)
+  ansible.builtin.apt_repository:
+    repo: "deb {{ pve_repository.uris }} {{ pve_repository.suites }} {{ pve_repository.components }}"
+    state: present
+    filename: proxmox
+  register: _pve_repo
+  when: ansible_distribution_major_version | int < 12
```
lae: I mentioned in the thread I linked that older Debians supposedly support deb822 repository format, so we should just be able to get rid of this task entirely. But it's up to you if you want to verify that and update this file accordingly. That said, the reason why I suggested translating `pve_repository_line` to `pve_repository` was so that we could use `pve_repository_line` in a deb822 repository configuration task. Parsing it out and then rebuilding the line here defeats the purpose, I think. (it lets new users use `pve_repository` against older PVE versions but I honestly doubt new users wouldn't just use Debian 12 from the start...)

tasks/repositories.yml
```diff
@@ -0,0 +1,103 @@
...
+
+- name: Ensure Ceph repository (PVE 8 and below) is compatible with new filename (PVE 9)
+  block:
+    - name: Check if old Ceph repository file exists
+      ansible.builtin.stat:
+        path: /etc/apt/sources.list.d/ceph.list
+      register: _pve_ceph_repo_old_file
+    - name: Rename old Ceph repository file to new filename
```
lae: [according to documentation](https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/command_module.html), this block can be simplified using the `removes` and `creates` attributes.

lae (changes requested):
Alright, got around to reviewing in more detail. See comments.

tasks/repositories.yml
PendaGTP: `components` can also be provided as elements string on a single line.

these two formats are equivalent:

```yml
# pve_repository_line: "deb http://download.proxmox.com/debian/pve {{ ansible_distribution_release }} pve-no-subscription test"

components: pve-no-subscription test
components:
  - pve-no-subscription
  - test
```
so I don’t think any change is required.

tasks/repositories.yml
PendaGTP: based on https://wiki.debian.org/SourcesList, it's indicated:

> Supported since (at least) [bullseye](https://wiki.debian.org/DebianBullseye), recommended since [trixie](https://wiki.debian.org/DebianTrixie)

so, we could normaly use ` < 11` instead of `< 12`

i keep this "logic" since on the repo there is [vars/debian-buster.yml](https://github.com/lae/ansible-role-proxmox/blob/15e20274331b98ef7c424d77d9ff8abe0c5a91d3/vars/debian-buster.yml), I supposed that we need to maintain compatibility with `buster`

idk which versions of Promox and Debian this role is supposed to support. I think dropping support for Proxmox versions older than 8 would be a good move
-> much easier to fix/improve other some things, without worrying about compatibility beetween 6,7,8,9 (+debian)
-> what do you think?

tasks/repositories.yml
lae: Okay, that's good, so the defaults can be left as is. But from a code quality perspective even if both are valid then the `join` is a bit redundant here.

tasks/repositories.yml
PendaGTP: nice catch, I didn't know about that

tasks/repositories.yml
PendaGTP: yup that's right!

tasks/repositories.yml
lae: Yeah, PVE 6 and 7 are EOL so we do want to remove them from the role. I didn't even realize we were keeping them around. We typically did remove EOL releases when adding new PVE versions (see #155, #33) but I guess I forgot to make sure they were removed when reviewing the PRs for adding PVE 8 and 9...

I should really draft a checklist for new releases.

PendaGTP: Thanks, that's good news! I'm working on other fixes/improvements that were a llittle bit burnning my head, trying to make it compatible with versions 6 and 7.

lae (approved):
Okay, I think we're good. Waiting for the vagrant test just to make sure it goes through these tasks fine.

PendaGTP:
Thanks for your time and helping me

PR #348, opened 2026-01-08
Add 'ceph-volume' to default installation packages for ceph

mabezi:
Well, there is an entire module called pve_ceph_volume, but the package is not installed automatically. (ceph-volume is also not installed by apt as a dependency in any form.)

### DIFF
--- a/tasks/identify_needed_packages.yml
+++ b/tasks/identify_needed_packages.yml
@@ -14,7 +14,7 @@
 
 - name: Stage Ceph packages if Ceph is enabled
   ansible.builtin.set_fact:
-    _pve_install_packages: "{{ _pve_install_packages | union(['ceph', 'ceph-common', 'ceph-mds', 'ceph-fuse', 'gdisk']) }}"
+    _pve_install_packages: "{{ _pve_install_packages | union(['ceph', 'ceph-common', 'ceph-mds', 'ceph-fuse', 'ceph-volume', 'gdisk']) }}"
   when: "pve_ceph_enabled | bool"
 
 - name: Stage any extra packages the user has specified


lae (changes requested):
Looks fine, however the package doesn't exist/wasn't necessary in bookworm (Debian 12/PVE 8):

https://packages.debian.org/bookworm/ceph-volume
https://packages.debian.org/trixie/ceph-volume

So there needs to be another conditional for that I think?

mabezi:
> Looks fine, however the package doesn't exist/wasn't necessary in bookworm (Debian 12/PVE 8):
> 
> https://packages.debian.org/bookworm/ceph-volume
> https://packages.debian.org/trixie/ceph-volume
> 
> So there needs to be another conditional for that I think?

This package is not included in the Debian Repositories, but in the PVE Ceph Repository.

lae:
I mean, it exists in the Trixie (Debian 13) repository in the second link I shared.

But you're right, Ceph packages should be getting pulled from PVE's Ceph repositories. That kind of makes things more complicated...I am going off the assumption that `ceph-volume` was split out as its own package at a certain Ceph version. So that might mean that not all of the PVE maintained repositories have the package?

mabezi:
I think it came with Ceph 15. (Which means PVE 6, right?)

But I'm not 100% sure right now

lae:
ok, I checked, `ceph-volume` exists in all active `no-subscription` ceph repositories for Debian 12:

http://download.proxmox.com/debian/ceph-quincy/dists/bookworm/no-subscription/binary-amd64/ceph-volume_17.2.8-pve1_all.deb
http://download.proxmox.com/debian/ceph-reef/dists/bookworm/no-subscription/binary-amd64/ceph-volume_18.2.7-pve1_all.deb
http://download.proxmox.com/debian/ceph-squid/dists/bookworm/no-subscription/binary-amd64/ceph-volume_19.2.3-1~bpo12%2B1_all.deb

as for when it was introduced, just going off the PVE package repositories it looks like Ceph 16 (Pacific, doesn't exist here http://download.proxmox.com/debian/ceph-pacific/dists/bullseye/main/binary-amd64/). But no Bookworm repo there or earlier, suggesting it was introduced in PVE 8, so I guess we can ignore them.

lae (approved):
:package:

PR #349, opened 2026-01-12
remove non existing kernel module (pve>=8)

PendaGTP:
The kernel module `vfio_virqfd` is no longer a module in Proxmox >=8

Loading it, throw an error at boot `modprobe: FATAL: Module vfio_virqfd not found`

Nothing breaking but just noisy


lae (approved):
Thanks~

PR #353, opened 2026-02-23
Fix #352: Disable logging of secrets by default

### LINKED ISSUE #352: Contents of pve_ssl_certificate and pve_ssl_private_key are leaked in Ansible run
Because Ansible outputs item values when using `with_items`, the contents of the variables `pve_ssl_certificate` and `pve_ssl_private_key` are shown in the `ansible-playbook` output:

```
TASK [lae.proxmox : Copy PVE SSL certificate chain and key to /etc/ssl] ******
changed: [pve.example.com] => (item={'dest': '/etc/ssl/pveproxy-ssl.key', 'content': '-----BEGIN PRIVATE KEY-----\nxxxxxx\n-----END PRIVATE KEY-----'})
changed: [pve.example.com] => (item={'dest': '/etc/ssl/pveproxy-ssl.pem', 'content': '-----BEGIN CERTIFICATE-----\nxxxxxx\n-----END CERTIFICATE-----'})
```

lae:
Fixes #352.

There are probably some people who might consider this a breaking change but I'd say the majority would not.

- **fix: set no_log on SSL configuration task**
- **feat: update pve_no_log to true (secure default)**

### DIFF
--- a/README.md
+++ b/README.md
@@ -436,7 +436,7 @@ pve_storages: [] # List of storages to manage in PVE. See section on Storage Man
 pve_metric_servers: [] # List of metric servers to configure in PVE.
 pve_datacenter_cfg: {} # Dictionary to configure the PVE datacenter.cfg config file.
 pve_domains_cfg: [] # List of realms to use as authentication sources in the PVE domains.cfg config file.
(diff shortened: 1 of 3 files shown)


thystips (approved):
Best version of my #354

PR #354, opened 2026-03-02
refactor: Add no_log option to SSL config task

thystips:
Display certificates content in Ansible stdout can be a serious security issue when run in CI/CD


thystips:
I hadn't see #353

